Issue #90, opened 2020-04-22
Pie.java:111-113: Extract the logic in switch into...


0pdd:
The puzzle `76-fe362791` from #76 has to be resolved: 

https://github.com/artipie/artipie/blob/95e04444ef222129f0e3dfdde67fb65ee884bcf6/src/main/java/com/artipie/Pie.java#L111-L113

The puzzle was created by @g4s8 on 22-Apr-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
The job [#90](https://github.com/artipie/artipie/issues/90) assigned to @hdouss[/z](https://www.0crat.com/u/hdouss), here is [why](https://www.0crat.com/footprint/CT2E6TK9B/2ad88dba-5acf-459f-a316-02435cb3b637); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be a monetary reward for this job

0pdd:
The puzzle `76-fe362791` has disappeared from the source code, that's why I closed this issue.

0pdd:
@0pdd 2 puzzles [#109](https://github.com/artipie/artipie/issues/109), [#110](https://github.com/artipie/artipie/issues/110) are still not solved.

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @hdouss[/z](https://www.0crat.com/u/hdouss), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

sereshqua:
@0crat quality good

0pdd:
@0pdd the puzzle [#110](https://github.com/artipie/artipie/issues/110) is still not solved; solved: [#109](https://github.com/artipie/artipie/issues/109).

[...]

PR #100, opened 2020-04-25
#90 - SliceFromConfig class

HDouss:
Resolving issue #90 
Added SliceFromConfig class
Added puzzle to test the new class

### DIFF
--- a/src/main/java/com/artipie/Pie.java
+++ b/src/main/java/com/artipie/Pie.java
@@ -91,7 +91,7 @@ public final class Pie implements Slice {
                     storage -> storage.value(new Key.From(String.format("%s.yaml", repo)))
                 )
                 .thenApply(content -> new RepoConfig(this.vertx, content))
-                .thenCompose(cfg -> CompletableFuture.completedStage(new SliceFromConfig(cfg)))
+                .thenApply(cfg -> new SliceFromConfig(cfg))
         ).response(line, headers, body);
     }
 }
--- a/src/main/java/com/artipie/Pie.java
+++ b/src/main/java/com/artipie/Pie.java
@@ -25,34 +25,24 @@
 package com.artipie;
 
 import com.artipie.asto.Key;
-import com.artipie.composer.http.PhpComposer;
-import com.artipie.files.FilesSlice;
-import com.artipie.gem.GemSlice;
 import com.artipie.http.Response;
 import com.artipie.http.Slice;
 import com.artipie.http.async.AsyncSlice;
 import com.artipie.http.rq.RequestLineFrom;
 import com.artipie.http.rs.RsStatus;
 import com.artipie.http.rs.RsWithStatus;
-import com.artipie.maven.http.MavenSlice;
-import com.artipie.npm.Npm;
-import com.artipie.npm.http.NpmSlice;
-import com.artipie.rpm.http.RpmSlice;
 import com.jcabi.log.Logger;
 import io.vertx.reactivex.core.Vertx;
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.function.Function;
 import org.cactoos.scalar.Unchecked;
 import org.reactivestreams.Publisher;
 
 /**
  * Pie of slices.
  * @since 1.0
- * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  * @checkstyle ReturnCountCheck (500 lines)
  */
 public final class Pie implements Slice {
@@ -101,54 +91,7 @@ public final class Pie implements Slice {
                     storage -> storage.value(new Key.From(String.format("%s.yaml", repo)))
                 )
                 .thenApply(content -> new RepoConfig(this.vertx, content))
-                .thenCompose(Pie::sliceForConfig)
+                .thenCompose(cfg -> CompletableFuture.completedStage(new SliceFromConfig(cfg)))
         ).response(line, headers, body);
     }
-
-    /**
-     * Find a slice implementation for config.
-     * @param cfg Repository config
-     * @return Async slice
-     * @todo #76:30min Extract the logic in switch into separate class.
-     *  It can be named like `SliceFromConfig`: it implements Slice interface
-     *  and behaves as a factory by creating `Slice` instance for configuration.
-     */
-    private static CompletionStage<Slice> sliceForConfig(final RepoConfig cfg) {
-        return cfg.type().thenCombine(
-            cfg.storage(),
-            (type, storage) -> {
-                final CompletionStage<Slice> slice;
-                switch (type) {
-                    case "file":
-                        slice = CompletableFuture.completedStage(new FilesSlice(storage));
-                        break;
-                    case "npm":
-                        slice = CompletableFuture.completedStage(
-                            new NpmSlice(new Npm(storage), storage)
-                        );
-                        break;
-                    case "gem":
-                        slice = CompletableFuture.completedStage(new GemSlice(storage));
-                        break;
-                    case "rpm":
-                        slice = CompletableFuture.completedStage(new RpmSlice(storage));
-                        break;
-                    case "php":
-                        slice = cfg.path().thenApply(
-                            path -> new PhpComposer(path, storage)
-                        );
-                        break;
-                    case "maven":
-                        slice = CompletableFuture.completedStage(new MavenSlice(storage));
-                        break;
-                    default:
-                        throw new IllegalStateException(
-                            String.format("Unsupported repository type '%s'", type)
-                        );
-                }
-                return slice;
-            }
-        ).thenCompose(Function.identity());
-    }
 }
-
--- a/src/main/java/com/artipie/SliceFromConfig.java
+++ b/src/main/java/com/artipie/SliceFromConfig.java
@@ -86,6 +86,10 @@ public final class SliceFromConfig implements Slice {
      * @return Slice
      * @throws ExecutionException If error getting the slice
      * @throws InterruptedException If error getting the slice
+     * @todo #90:30min This method still needs more refactoring. First, we should extract
+     *  PhpComposer Slice construction to a private method to make it more readable. And then,
+     *  we should test if the type exist in the constructed map. If the type does not exist,
+     *  we should throw an IllegalStateException with the message "Unsupported repository type '%s'"
      */
     private static Slice build(final RepoConfig cfg) throws InterruptedException,
         ExecutionException {
--- a/src/main/java/com/artipie/SliceFromConfig.java
+++ b/src/main/java/com/artipie/SliceFromConfig.java
@@ -37,6 +37,9 @@ import com.jcabi.log.Logger;
 import java.nio.ByteBuffer;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+import org.cactoos.map.MapEntry;
+import org.cactoos.map.MapOf;
 import org.reactivestreams.Publisher;
 
 /**
@@ -44,6 +47,7 @@ import org.reactivestreams.Publisher;
  * @since 0.1.4
  * @todo #90:30min We still don't have tests for Pie. But now that this class was extracted, we have
  *  a more cohesive class that could be tested. Write unit tests for SliceFromConfig class.
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  */
 public final class SliceFromConfig implements Slice {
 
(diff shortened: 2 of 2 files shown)


0crat:
This pull request [#100](https://github.com/artipie/artipie/issues/100) is assigned to @vatavuk[/z](https://www.0crat.com/u/vatavuk), here is [why](https://www.0crat.com/footprint/CT2E6TK9B/1d6fe425-da76-4b34-be4f-6495b493afba); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @g4s8[/z](https://www.0crat.com/u/g4s8) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be a monetary reward for this job

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,130 @@
...
+                        slice = new GemSlice(storage);
+                        break;
+                    case "rpm":
+                        slice = new RpmSlice(storage);
+                        break;
+                    case "php":
+                        try {
+                            slice = cfg.path().thenApply(
+                                path -> new PhpComposer(path, storage)
+                            ).toCompletableFuture().get();
```
Vatavuk, line 109: @HDouss I dont think we should execute the request here, it should return `CompletitionStage`

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.concurrent.ExecutionException;
+import org.reactivestreams.Publisher;
+
+/**
+ * Slice from repo config.
+ * @since 0.1.4
+ * @todo #90:30min We still don't have tests for Pie. But now that this class was extracted, we have
+ *  a more cohesive class that could be tested. Write unit tests for SliceFromConfig class.
+ */
+public final class SliceFromConfig implements Slice {
```
Vatavuk, line 48: @HDouss I don't think that this class can implement `Slice` since it has to build `CompletionStage<Slice>` in order to be async.

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,130 @@
...
+     * @throws ExecutionException If error getting the slice
+     * @throws InterruptedException If error getting the slice
+     */
+    private static Slice build(final RepoConfig cfg) throws InterruptedException,
+        ExecutionException {
+        return cfg.type().thenCombine(
+            cfg.storage(),
+            (type, storage) -> {
+                final Slice slice;
+                switch (type) {
```
Vatavuk, line 92: @HDouss I believe that we can use `Map` instead of ugly switch/case.

src/main/java/com/artipie/Pie.java
```diff
@@ -101,54 +91,7 @@ public Response response(final String line, final Iterable<Map.Entry<String, Str
                     storage -> storage.value(new Key.From(String.format("%s.yaml", repo)))
                 )
                 .thenApply(content -> new RepoConfig(this.vertx, content))
-                .thenCompose(Pie::sliceForConfig)
+                .thenCompose(cfg -> CompletableFuture.completedStage(new SliceFromConfig(cfg)))
```
Vatavuk, line 94: @HDouss `SliceFromConfig` should build `CompletionStage<Slice>` in order to be async and we should remove `completedStage` here

src/main/java/com/artipie/SliceFromConfig.java
HDouss: @Vatavuk it is not executing the request, it is resolving config.path and build the slice. This method is returning a slice

src/main/java/com/artipie/SliceFromConfig.java
HDouss: @Vatavuk That's what the puzzle (issue description) tells. Anyway, I defered the slice building and response acting to the response method so it would be async.

src/main/java/com/artipie/SliceFromConfig.java
HDouss: @Vatavuk If we use Map, what could be the values? Classes that we instantiate by reflection afterwards? That would be ugly too...may be uglier.

src/main/java/com/artipie/Pie.java
HDouss: @Vatavuk That what the description tell, anyway the constructor is too fast, it is even faster than the construction of the `CompletionStage<Slice>` that was made by `Pie::sliceForConfig`. The actual logic is defered in the `response` medthod.

src/main/java/com/artipie/SliceFromConfig.java
Vatavuk: @HDouss We are fetching yaml content from a repo in an asynchronous manner. The code you've wrote will be synchronous.

src/main/java/com/artipie/SliceFromConfig.java
Vatavuk: @HDouss it is not deferred because of instantiation of `php` slice which needs to fetch path from the repo. The code for `php` instantiation executes `.toCompletableFuture().get()`.

src/main/java/com/artipie/SliceFromConfig.java
Vatavuk: @HDouss there are no reflections. Simple `HashMap`
```java
new MapOf(
  new MapEntry("file", (cfg) -> new FilesSlice(storage))),
  new MapEntry("npm" (cfg) -> new NpmSlice(new Npm(storage), storage))
  ...
)
```
The map should probably return `CompletionStage` instead of `Slice` so that the code remain async.

src/main/java/com/artipie/Pie.java
Vatavuk: @HDouss If you make `SliceFromConfig` async we don't need `completedStage`. If you think that `SliceFromConfig` doesn't need to be async please say why, because this wasn't the original behaviour.

Vatavuk: @HDouss I know that the issue says that we should implement `Slice` just to make code more reusable, a refactoring. But I don't see how can we do this unless we change the behaviour.

Vatavuk: @g4s8 maybe little help here. The purpose of this feature is to break the class into the smaller chunks. It was written with refactoring on the mind and I believe it should not change any behaviour. That being said this PR instantiates `Slice` objects synchronously as opposed to before. Is this ok or should we revert?

src/main/java/com/artipie/SliceFromConfig.java
HDouss: @Vatavuk Actually it is not really synchronous. Because all this method is called when the `response` method of the slice is called. What is really synchronous is the call to the constructor of `SliceFromConfig` which is much less than the older switch/case logic (that was actually synchronous)

src/main/java/com/artipie/SliceFromConfig.java
HDouss: @Vatavuk That still not happen until `SliceFromConfig.response` is called, which is deferred.

src/main/java/com/artipie/Pie.java
HDouss: @Vatavuk Please see my replies earlier. It is true that it is a slice, and not a `CompletionStage<Slice>`, but still what is really synchronous is merely the call to the constructor. ~~And anyway, how to "make SliceFromConfig async" without making it a utility class?~~

src/main/java/com/artipie/SliceFromConfig.java
HDouss: @Vatavuk Thanks for your explanation. Fixed!

src/main/java/com/artipie/SliceFromConfig.java
Vatavuk: @HDouss you are right, I didn't catch the execution of the methotd itself

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,135 @@
...
+                    ),
+                    new MapEntry<String, Function<RepoConfig, Slice>>(
+                        "gem", config -> new GemSlice(storage)
+                    ),
+                    new MapEntry<String, Function<RepoConfig, Slice>>(
+                        "rpm", config -> new RpmSlice(storage)
+                    ),
+                    new MapEntry<String, Function<RepoConfig, Slice>>(
+                        "php",
+                        config -> {
```
Vatavuk, line 110: @HDouss extract this to a private method to make it more readable

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,135 @@
...
+                                    SliceFromConfig.class, "Exception getting PhpComposer"
+                                );
+                                throw new IllegalArgumentException(ex);
+                            }
+                        }
+                    ),
+                    new MapEntry<String, Function<RepoConfig, Slice>>(
+                        "maven", config -> new MavenSlice(storage)
+                    )
+                ).get(type).apply(cfg);
```
Vatavuk, line 131: @HDouss we should check if the type exists in the map and throw exception if it doesnt

HDouss:
@Vatavuk I added a puzzle for the two remaining comments. I largely exceeded 30 min here.

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,139 @@
...
+     * @param cfg Repository config
+     * @return Slice
+     * @throws ExecutionException If error getting the slice
+     * @throws InterruptedException If error getting the slice
+     * @todo #90:30min This method still needs more refactoring. First, we should extract
+     *  PhpComposer Slice construction to a private method to make it more readable. And then,
+     *  we should test if the type exist in the constructed map. If the type does not exist,
+     *  we should throw an IllegalStateException with the message "Unsupported repository type '%s'"
+     */
+    private static Slice build(final RepoConfig cfg) throws InterruptedException,
```
g4s8, line 94: @HDouss why do we have `InterruptedException` here? Does this method perform any blocking operation? We can't allow thread blocking for `Slice` implementations.

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,139 @@
...
+                    ),
+                    new MapEntry<String, Function<RepoConfig, Slice>>(
+                        "php",
+                        config -> {
+                            try {
+                                return config.path().thenApply(
+                                    path -> new PhpComposer(path, storage)
+                                ).toCompletableFuture().get();
+                            } catch (final InterruptedException ex) {
+                                Logger.error(
```
g4s8, line 120: @HDouss same here: why `InterruptedException`? Please don't use `get()` method anywhere except test methods.

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,139 @@
...
+                                throw new IllegalArgumentException(ex);
+                            }
+                        }
+                    ),
+                    new MapEntry<String, Function<RepoConfig, Slice>>(
+                        "maven", config -> new MavenSlice(storage)
+                    )
+                ).get(type).apply(cfg);
+            }
+        ).toCompletableFuture().get();
```
g4s8, line 137: @HDouss we can't block HTTP thread with blocking `get()` call

HDouss:
@g4s8 The synchronous operations was in the response method. I fixed that but the only way to do it is to wrap the response again in an `asyncslice.response`.

src/main/java/com/artipie/Pie.java
g4s8, line 94: @HDouss maybe it would be better to use `.thenApply(cng -> new SliceFromConfig(cfg))` here?

HDouss: @g4s8 Fixed!

HDouss: @g4s8 I don't see the second minor comment..

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -0,0 +1,118 @@
...
+     * @param config Repo config
+     */
+    public SliceFromConfig(final RepoConfig config) {
+        this.config = config;
+    }
+
+    @Override
+    public Response response(final String line, final Iterable<Map.Entry<String, String>> headers,
+        final Publisher<ByteBuffer> body) {
+        return new AsyncSlice(
```
g4s8, line 71: @HDouss instead of creating `AsyncSlice` instance in `response` method of `Slice` on each request you can use `Slice.Wrap` decorator for wrapping this logic:
```java
class SliceFromConfig extends Slice.Wrap {
  public SliceFromConfig(RepoConfig config) {
    super(
      new AsyncSlice(SliceFromConfig.build(config))
    );
  }
}
```

HDouss: @g4s8 I don't think it is a good idea. First, that will make the call to `SliceFromConfig.build` synchronous with the SliceFromConfig construction, so with Pie.response method. Second, for the requests coming from Pie, it is anyway building a new SliceFromConfig on each request, it is even building a new wrapping AsyncSlice.

g4s8: @HDouss but `SliceFromConfig.build(config)` is returning `CompletionStage` so it's not synchronous, right? Then actual creating of `Slice` will be deferred until next request. I didn't get your second point, but you are right: we're building new `Slice` for each request (we may add some caching later), it's correct.
The problem with current implementation is only in the design: you are creating new instance of same level in method, but it can be moved to constructor. Creating new concrete implementation in method increases class coupling, but constructor may accept interface types, so coupling will be low.

HDouss: @g4s8 What I meant is that the call to build is synchronous. But you are right, now with the last commit modifications it is not synchronous anymore. The second point is about the usage by Pie. Anyway fixed. Please see last commit.

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @vatavuk[/z](https://www.0crat.com/u/vatavuk), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

sereshqua:
@0crat quality good

PR #163, opened 2020-05-21
remove sources from master and fix REST path in benchmarks

brastak:
Remove redundant files in gh-pages branch and fix REST call


0crat:
Job `gh:artipie/artipie#163` is not assigned, can't get performer

Issue #307, opened 2020-07-10
Add GitHub magic files


g4s8:
Add contributing policy, issues and PR templates and other GitHub magic files.

PR #308, opened 2020-07-10
GitHub files

g4s8:
#307 - added GitHub files for the repo


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,5 @@
+# Contributor Code of Conduct
+
+This project adheres to No Code of Conduct.  We are all adults.  We accept anyone's contributions.  Nothing else matters.
```
Sammers21, line 3: A useless file. Remove it.

CODE_STYLE.md
```diff
@@ -1,22 +0,0 @@
...
-```
-
-2. Avoid adding reason to assertions in unit tests with single assertion:
-```java
-// use
-MatcherAssert.assertThat(target, matcher);
-
-// don't use
-MatcherAssert.assertThat("Some reason", target, matcher);
-```
```
Sammers21, line 22: I think this file is not related to the pull request

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
+To contribute to Artipie project you need JDK-11 and Maven 3.2+.
```
Sammers21, line 1: No reference link is provided.

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+Some integration tests requires Docker to be installed.
+
+
+## How to contribute
+
+Fork the repository, make changes, and send us a
+[pull request](https://www.yegor256.com/2014/04/15/github-guidelines.html). We will review
```
Sammers21, line 8: Invalid link for reference. Use this one: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests

CODE_STYLE.md
g4s8: @Sammers21 `CODE_STYLE.md` is not a standard GitHub file, so it was moved to `CONTRIBUTING.md` which is the standard file and GitHub displays a reference when submitting pull request.

CONTRIBUTING.md
g4s8: @Sammers21 reference to what? I suppose contributors are able to install Java and Maven by themselves, also there are no standard links for that - each OS distribution has its own package for Java and Maven.

CODE_OF_CONDUCT.md
g4s8: @Sammers21 GitHub doesn't think so, it asks to add such file
scrot-2020-07-13-12:08:54

CONTRIBUTING.md
g4s8: @Sammers21 the link you have provided just explains what pull request is but not how to use it properly.

CONTRIBUTING.md
Sammers21: @g4s8, for jdk 11 you can refer to https://adoptopenjdk.net/ or https://jdk.java.net/11/. For maven, you can refer to https://maven.apache.org/

CONTRIBUTING.md
Sammers21: The only mention of pull requests in the article refers to the exact Github link I sent above:
image

CODE_OF_CONDUCT.md
Sammers21: @g4s8, A code of conduct is a set of rules outlining the norms, rules, and responsibilities or proper practices of an individual party or an organisation. We work with zerocrat and have specific rules. Mention them in the file. Now the file does not bring any value.

g4s8: @Sammers21 we don't work by Zerocracy rules, only users from 0crat.com do. But they are aware about these rules without this file.

0crat:
Job `gh:artipie/artipie#308` is not assigned, can't get performer

Issue #285, opened 2020-07-08
Use Github tokens for central.artipie.com users


g4s8:
Use GitHub personal tokens as HTTP basic authentication for central users, user generates token in personal settings, uses it as password for authentication and uses special username like `github.com/username`. Artipie authenticates user via GitHub API by token and validates username.

Sammers21:
0 references to the official documentation. I suggest you add them

0pdd:
@g4s8 2 puzzles [#337](https://github.com/artipie/artipie/issues/337), [#339](https://github.com/artipie/artipie/issues/339) are still not solved.

g4s8:
Done

0pdd:
@g4s8 the puzzle [#337](https://github.com/artipie/artipie/issues/337) is still not solved; solved: [#339](https://github.com/artipie/artipie/issues/339).

PR #318, opened 2020-07-13
Authentication by github tokens

g4s8:
#285 - authentication by GitHub tokens:
 - added GitHub auth provider
 - added chained auth provider to compose different providers
 - added cached auth decorator to avoid fetching GitHub API too frequently
 - added puzzles for cache cleanup and auth settings configuration


src/test/java/com/artipie/auth/CachedAuthTest.java
```diff
@@ -0,0 +1,75 @@
...
+    void callsOriginOnlyOnce() {
+        final String username = "usr";
+        final AtomicInteger counter = new AtomicInteger();
+        final CachedAuth target = new CachedAuth(
+            (usr, pass) -> Optional.of(String.format("%s-%d", usr, counter.incrementAndGet())),
+            new ConcurrentHashMap<>()
+        );
+        final String expected = "usr-1";
+        MatcherAssert.assertThat(
+            "Wrong user on first cache call",
```
olenagerasimova, line 65: @g4s8 hm, assertion message usually (in EO style code) specify that is expected, not what was wrong. I'm not sure this is some kind of rule though and hamcrest does not specify what we should put in the message...

src/test/java/com/artipie/auth/CachedAuthTest.java
```diff
@@ -0,0 +1,75 @@
...
+            target.user(username, "").orElseThrow(),
+            new IsEqual<>(expected)
+        );
+        MatcherAssert.assertThat(
+            "Wrong user on second cache call",
```
olenagerasimova, line 70: @g4s8 same here

src/main/java/com/artipie/auth/GithubAuth.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.IOException;
+import java.util.Locale;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * GitHub authentication uses prefixed by provider name username and personal access token
```
Sammers21, line 37: 1. No reference to the official docs.
2. what does "prefixed by provider name username" mean?

src/main/java/com/artipie/auth/ChainedAuth.java
```diff
@@ -0,0 +1,72 @@
...
+package com.artipie.auth;
+
+import com.artipie.http.auth.Authentication;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Chained authentication provider, composed by multiple
```
Sammers21, line 33: The strangest general description I saw in my life. Can you explain for which purpose you plan to use it?

pom.xml
```diff
@@ -115,6 +115,11 @@ SOFTWARE.
       <artifactId>asto</artifactId>
       <version>0.23.1</version>
     </dependency>
+    <dependency>
```
Sammers21, line 118: The library is synchronous, it would not work well. You can make HTTP requests to Github with Vert.x HTTP client.

src/main/java/com/artipie/YamlSettings.java
```diff
@@ -45,6 +47,11 @@
  * Settings built from YAML.
  *
  * @since 0.1
+ * @todo #285:30min Settings configuration for GitHub auth.
```
Sammers21, line 50: I can't understand the description. Can you explain what already done and what is planned to do in the puzzle?

src/main/java/com/artipie/auth/CachedAuth.java
```diff
@@ -0,0 +1,83 @@
...
+ * Cached authentication decorator.
+ * <p>
+ * It remembers the result of decorated authentication provider and returns it
+ * instead of calling origin authentication.
+ * </p>
+ * @since 0.10
+ * @todo #285:30min Specify expiration time configuration.
+ *  Instead of using scheduled executor to clean-up all cache map, use
+ *  some configuration to clean-up only expired items, e.g. if token was not accessed for
+ *  X minutes, then remove only this token.
```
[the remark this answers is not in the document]
Sammers21: @olenagerasimova, a good point.

src/main/java/com/artipie/auth/GithubAuth.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public Optional<String> user(final String username, final String password) {
+        Optional<String> result = Optional.empty();
+        final Matcher matcher = GithubAuth.PTN_NAME.matcher(username);
+        if (matcher.matches()) {
+            final String login = this.github.apply(password).toLowerCase(Locale.US);
+            if (
+                Objects.equals(login, matcher.group(1).toLowerCase(Locale.US))
+            ) {
```
Sammers21, line 87: The if statement looks ugly

src/test/java/com/artipie/auth/CachedAuthTest.java
g4s8: @olenagerasimova assertion description is shown only when assertion fails, so it looks logical to say what is wrong. The example: `Wrong user on first cache call: expected "usr", but was "other"`.

src/main/java/com/artipie/auth/ChainedAuth.java
g4s8: @Sammers21 can you explain what exactly is not clear here? It says `composed by multiple authentication providers invoked by user specified order.`.

pom.xml
g4s8: @Sammers21 it doesn't make any sense here: authentication is a blocking operation in any case, we can't continue processing the request without performing synchronous call and authenticating the user.

src/main/java/com/artipie/auth/GithubAuth.java
g4s8: @Sammers21 it's not constructive. I've either suggest an improvement or say what exactly is wrong.

g4s8:
@olenagerasimova @Sammers21 thanks, fixed and answered: https://github.com/artipie/artipie/pull/318/commits/63f377d39cc64aa1aa886cacd32cfa6701888e65

pom.xml
Sammers21: @g4s8, it is a blocking process for a user, not necessary for the application.

g4s8: @Sammers21 I don't see any benefits using reactive client here: we still need to wait for response in a blocking way, so the performance of this call will be the same.

src/main/java/com/artipie/auth/ChainedAuth.java
Sammers21: @g4s8, I got the meaning, but it is still hard for me to imagine. An example would be helpful.

src/main/java/com/artipie/auth/GithubAuth.java
Sammers21: @g4s8, beauty can't be constructive or non-construtive. It is a personal matter

src/main/java/com/artipie/auth/GithubAuth.java
```diff
@@ -0,0 +1,95 @@
...
+                } catch (final IOException unauthorized) {
+                    return "";
+                }
+            }
+        );
+    }
+
+    /**
+     * Primary constructor.
+     * @param github Github username resolver
```
Sammers21, line 75: Resoloves tokens to username? Right? can you specify it?

pom.xml
Sammers21: @g4s8, i guess that is because interface design is bad. The user method should be returning `Maybe<String>` or `Single<Optional<String>>` or `CompletableFuture<Optional<String>>`.

g4s8:
@Sammers21 thanks, fixed some comments

PR #379, opened 2020-07-22
#371 - Trimmed image names in Docker repositories

### LINKED ISSUE #371: Trim repo name from Docker image name when stored
Docker adapter release `0.3.1` supports `TrimmedDocker` that cuts specified prefix in repo names when they are stored or loaded from storage. `TrimmedDocker` should be used to wrap underlying `Docker` instance both in for `docker` and `docker-proxy` repo types with prefix being repo name.

olegmoz:
Closes #371 
Trimmed image names in Docker repositories


olegmoz:
@victornoel I absolutely agree that tests are missing, but design of classes does not allow to test the changes with unit tests. In tests I should create a `Slice` from YAML config and see if saved data goes to proper path in `Storage` or requests are sent to proper URL in case of proxy. That's an integration test and there is no infrastructure ready for them right now, no any tests for `SliceFromConfig` or integration tests for whole application. So I have considered verifying the changes with tests, but the task seems to be too massive

victornoel:
@g4s8 @olenagerasimova should we add a todo for adding more tests or tackling this "massive" task that @olegmoz talks about?

victornoel:
@olegmoz changing the design of classes so that tests can be written should be a good solution IMHO

g4s8:
> @g4s8 @olenagerasimova should we add a todo for adding more tests or tackling this "massive" task that @olegmoz talks about?

@victornoel I think we need additional task for integration tests in this module

0crat:
Job was finished in 39 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

0crat:
The project doesn't have enough funds, can't make a payment

PR #400, opened 2020-07-29
#399 - Everthing is allowed when no permissions in repo YAML

### LINKED ISSUE #399: Everything should be allowed when no `permissions` in config
If permissions are checked in a slice and no `permissions` mapping found in repo YAML then application crashes. It seems logical that everything should be allowed for any user if there are no `permissions` specified (e.g. by default)

olegmoz:
Closes #399 
Everthing is allowed when no permissions in repo YAML


0crat:
Job `gh:artipie/artipie#400` is not assigned, can't get performer

Issue #409, opened 2020-07-31
Request may hugns forever


g4s8:
It's reproduced one time, the request was not responded for 15 minutes. Let's handle this scenario and respond with `503` error if request was not respond with any head for 1 minute.

PR #412, opened 2020-07-31
Cancel request by timeout

g4s8:
#409 - wrapped `Pie` slices with `TimeoutSlice`

### DIFF
--- a/src/main/java/com/artipie/http/TimeoutSlice.java
+++ b/src/main/java/com/artipie/http/TimeoutSlice.java
@@ -1,0 +1,213 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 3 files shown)


src/main/java/com/artipie/http/TimeoutSlice.java
```diff
@@ -0,0 +1,213 @@
...
+            this.exec.schedule(
+                () -> {
+                    if (this.completed.compareAndSet(false, true)) {
+                        this.origin.accept(
+                            RsStatus.UNAVAILABLE, Headers.EMPTY,
+                            new Content.From(
+                                "Request cancelled by timeout".getBytes(StandardCharsets.UTF_8)
+                            )
+                        );
+                        future.complete(null);
```
olegmoz, line 205: @g4s8 `future` should complete only after `this.origin.accept` has been completed, please move `future.complete` invocation to `.thenRun()`

src/main/java/com/artipie/http/TimeoutSlice.java
```diff
@@ -0,0 +1,213 @@
...
+                    }
+                },
+                duration.getSeconds(), TimeUnit.SECONDS
```
olegmoz, line 208: @g4s8 I'd suggest to use milliseconds here as it is default precision in Java libraries

g4s8:
@olegmoz thanks, should be fixed now: https://github.com/artipie/artipie/pull/412/commits/79a58b6ead915bbbdee277a7bb757a0d94a8303f

0crat:
Job `gh:artipie/artipie#412` is not assigned, can't get performer

Issue #425, opened 2020-08-07
Publish traffic and size metrics and publish to the dashboard.


g4s8:
Collect metrics for traffic (upload and download) and repository sizes, and show these metrics on the dashboard.

0pdd:
@g4s8 2 puzzles [#446](https://github.com/artipie/artipie/issues/446), [#447](https://github.com/artipie/artipie/issues/447) are still not solved.

g4s8:
Done in #439

0pdd:
@g4s8 the puzzle [#446](https://github.com/artipie/artipie/issues/446) is still not solved; solved: [#447](https://github.com/artipie/artipie/issues/447).

PR #426, opened 2020-08-07
Collected traffic metrics from requests and responses

g4s8:
#425 - collected size of requests body publisher and response body publisher and increments traffic metrics.

### DIFF
--- a/src/main/java/com/artipie/http/TrafficMetricSlice.java
+++ b/src/main/java/com/artipie/http/TrafficMetricSlice.java
@@ -1,0 +1,238 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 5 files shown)


src/main/java/com/artipie/http/TrafficMetricSlice.java
```diff
@@ -0,0 +1,238 @@
...
+
+        /**
+         * Metrics to update.
+         */
+        private final Metrics metrics;
+
+        /**
+         * Metrics name to update.
+         */
+        private final String name;
```
olegmoz, line 161: @g4s8 you may pass `Counter` to this class instead of two fields

src/test/java/com/artipie/http/TrafficMetricSliceTest.java
```diff
@@ -0,0 +1,108 @@
...
+                }
+
+                @Override
+                public void gauges(final Map<String, Long> data) {
+                    if (data.containsKey(TrafficMetricSliceTest.KEY_REQ)) {
+                        request.set(data.get(TrafficMetricSliceTest.KEY_REQ));
+                    }
+                    if (data.containsKey(TrafficMetricSliceTest.KEY_RSP)) {
+                        response.set(data.get(TrafficMetricSliceTest.KEY_RSP));
+                    }
```
olegmoz, line 95: @g4s8 it seems that this code block is not needed and so may be removed. It would be even better to throw an exception here, because we do not expect the slice to update any gauges.

src/test/java/com/artipie/http/TrafficMetricSliceTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Test
+    void updatesMetrics() throws Exception {
+        final Metrics metrics = new InMemoryMetrics();
+        final int size = 100;
+        new TrafficMetricSlice((line, head, body) -> new RsWithBody(body), metrics)
+            .response(
+                new RequestLine(RqMethod.GET, "/foo").toString(),
+                Collections.emptyList(),
+                Flowable.just(ByteBuffer.wrap(new byte[size]))
```
olegmoz, line 67: @g4s8 I'd suggest to have different size for request and response, otherwise we may mix up response and request in metrics and never know

g4s8:
@olegmoz thanks, I've fixed all comment. Also, I extracted `TestMetricsOutput` from another test class to use it instead of anonymous class for output.

0crat:
Job `gh:artipie/artipie#426` is not assigned, can't get performer

Issue #432, opened 2020-08-10
Week#33


g4s8:
This is the plan for `#33` week of 2020:
 - [x] Validate Maven artifacts on upload: https://github.com/artipie/maven-adapter/issues/125
 - [x] HTTP client for files-proxy repository: https://github.com/artipie/files-adapter/issues/19
 - [x] HTTP client for Maven adapter: https://github.com/artipie/maven-adapter/issues/128
 - [x] HTTP client for Docker adapter: https://github.com/artipie/docker-adapter/issues/304
 - [x] Aggregate and show published metrics from storage if enabled #425
 - [x] Implement storage locking system: https://github.com/artipie/asto/issues/214
 - [ ] Publish client benchmarks with charts to GH pages: https://github.com/artipie/artipie/issues/431

@artipie/contributors FYI

0crat:
Job `gh:artipie/artipie#432` is not assigned, can't get performer

PR #440, opened 2020-08-13
Updated files proxy

g4s8:
#432 - updated files proxy using new HTTP client slice


0crat:
Job `gh:artipie/artipie#440` is not assigned, can't get performer

Issue #392, opened 2020-07-24
Health check endpoint


cambierr:
To be able to deploy artipie in Kubernetes environments (for instance, using helm charts), it would be a really nice things to benefit from a health check endpoint to be able to monitor the application state and trigger some self-healing & scaling if needed.

To be more precise, I'd suggest to take inspiration from gitlab for instance: https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html

Any though on this ? like what should be taken into account to report a 100% UP check ? storages availability ? correct configuration ?a per-repository heath check interface to be able to report a global healthiness view ? ...?

jeffzhu503:
Yes, the service needs to expose these end points for monitoring purpose, even without using Kubernetes. Copy from the gitlab link from the above.  
GET http://localhost/-/health
GET http://localhost/-/readiness
GET http://localhost/-/liveness

g4s8:
@cambierr thanks for reporting, we definitely need health-check endpoints for k8s.
> Any though on this ? like what should be taken into account to report a 100% UP check ? storages availability ? correct configuration ?a per-repository heath check interface to be able to report a global healthiness view ? ...?

Storages availability - we can check repository configuration storage availability for the health-check, but I'd not check repository storages, since it may take a while.

Correct configuration - it's not needed if we check "Storage availability", since configuration will be parsed and validated to read storage config

A per-repository heath check - I don't think it's a good idea for some reasons: 1) if one repository is broken, we shouldn't stop other repositories which may be used 2) it could be a long-running operation if we have a lot of repositories

@jeffzhu503 @cambierr so I'd go with first option: checking main storage availability (which includes main configuration validation), WDYT?

g4s8:
@cambierr @jeffzhu503 health check endpoint was implemented in #458 PR.

0crat:
Job `gh:artipie/artipie#392` is not assigned, can't get performer

PR #458, opened 2020-08-18
Added health endpoint

g4s8:
#392 - added health endpoint for primary storage, it returns `OK` status if check passed and `BAD_REQUEST` (it seems to be the standard error for health check errors), and JSON array response with verbose statuses for each check (only storage for now).

### DIFF
--- a/src/main/java/com/artipie/http/HealthSlice.java
+++ b/src/main/java/com/artipie/http/HealthSlice.java
@@ -1,0 +1,100 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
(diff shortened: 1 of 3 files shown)


src/main/java/com/artipie/http/HealthSlice.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import javax.json.Json;
+import org.reactivestreams.Publisher;
+
+/**
+ * Health check slice.
+ * <p>
+ * Returns JSON with verbose status checks,
+ * response status is {@code OK} if all status passed and {@code BAD_REQUEST} if any failed.
```
olegmoz, line 46: @g4s8 I am not sure if you are adapting to some existing health check protocol, but it does not look good to respond with `4xx` client error response. I think some `5xx` server error is more appropriate, maybe 503 Service Unavailable will make more sense?

olegmoz (changes requested):
@g4s8 thanks, please see one comment

g4s8:
@olegmoz thanks, should be fixed in https://github.com/artipie/artipie/pull/458/commits/9491048d93801030f668f47c6d78f20693f03c9d

olegmoz (approved):
@g4s8 thanks, I think it's good to merge

olegmoz (approved):

0crat:
Job `gh:artipie/artipie#458` is not assigned, can't get performer

Issue #444, opened 2020-08-14
Support /secutiry API of Artifactory


g4s8:
Support [Artifactory API](https://www.jfrog.com/confluence/display/rtf/artifactory+rest+api) at `/api/secutiry/users` and `/api/secutiry/permissions`: API clients should be able to create update and delete users and permissions.

olenagerasimova:
@g4s8 are we are going to support all requests to `/api/secutiry/users` endpoint? There are several variations:
1. GET `/api/security/users` to obtain users list
2. GET `/api/security/users/{userName}` to obtain user info
3. PUT `/api/security/users/{userName} ` to create/replace user
4. POST `/api/security/users/{userName}` to update user
5. DELETE `/api/security/users/{userName}` to delete user
4. GET `/api/security/encryptedPassword` Get User Encrypted Password
6. POST `/api/security/users/authorization/changePassword` to change password
7. Several methods to expire/unexpire user password, expire password for user list, expire passwords for all users

2-4 work with json body [format](https://www.jfrog.com/confluence/display/JFROG/Security+Configuration+JSON#SecurityConfigurationJSON-application/vnd.org.jfrog.artifactory.security.User+json) where `email` field is required, we do not store user's emails.

g4s8:
@olenagerasimova we can start with CRUD operations for users (`GET /api/security/users`, `GET /api/security/users/{userName}`, `PUT /api/security/users/{userName}`, `POST /api/security/users/{userName}` and `DELETE /api/security/users/{userName}`). All other `/users` operations seems to be not necessary.

olenagerasimova:
@g4s8 ok, please check user artifactory [payload](https://www.jfrog.com/confluence/display/JFROG/Security+Configuration+JSON) 
```
{
  - "name": "davids",
  + "email" : "davids@jfrog.com",
  + "password": "***" (write-only, never returned),
  - "admin": false (default),
  - "profileUpdatable": true (default),
  - "disableUIAccess" : true,
  - "internalPasswordDisabled": false (default),
  ! "lastLoggedIn": ISO8601 (yyyy-MM-dd'T'HH:mm:ss.SSSZ),
  ! "realm": "Internal",
  - "groups" : [ "deployers", "users" ],
  - "watchManager": false(default),
  - "policyManager": false(default)
}
``` 
fields with + are required, - optional, ! read only. Questions:
1. We do not have users' email to return - what should I do about it?
[...]

PR #467, opened 2020-08-20
#444 - Credentials interface introduced

olenagerasimova:
Part of #444 
Introduced `Credentials` interface to work with credentials and to be used in artifactory API for users CRUD operation, it's implementation, implemented method to read yaml from credentials file and used it in `GetUserSlice`.
Alos, simplified `YamlSettings` and its test.


src/main/java/com/artipie/Credentials.java
```diff
@@ -0,0 +1,103 @@
...
+import com.artipie.asto.Storage;
+import com.artipie.asto.rx.RxStorageWrapper;
+import hu.akarnokd.rxjava2.interop.SingleInterop;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * Artipie credentials.
+ * @since 0.9
+ */
+public interface Credentials {
```
olegmoz, line 38: @olenagerasimova looking at `add` and `remove` comments it seems that `Credentials` interface assumes that it's data is stored in YAML. Is it the case? If it's for YAML credentials only then I'd suggest to rename the interface accordingly and describe it in class comments. If it is supposed to handle credentials in other formats as well, then corrections should be made to methods' comments.

src/main/java/com/artipie/Credentials.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param username User to delete
+     * @return Completion remove action
+     */
+    CompletionStage<Void> remove(String username);
+
+    /**
+     * Credentials from main artipie config.
+     * @since 0.9
+     */
+    final class FromConfig implements Credentials {
```
olegmoz, line 65: @olenagerasimova looking at this class, I see that it's main feature is to load the YAML from `Storage`.  So I'd suggest to rename it to `FromStorageKey`, `FromStorageYaml` or something like that.

olegmoz (changes requested):
@olenagerasimova thanks, please see two comments about naming

src/main/java/com/artipie/Credentials.java
olenagerasimova: @olegmoz thanks, I've made `Credentials` more general: now there is not yaml, but users list, this way it also will be very convenient. Implementation I've called `FromStorageYaml`, as it works with yaml from storage.

olenagerasimova:
@olegmoz thanks, made some corrections, please, take a look

src/main/java/com/artipie/YamlSettings.java
```diff
@@ -86,7 +86,7 @@ public Storage storage() throws IOException {
     public CompletionStage<Authentication> auth() {
         return this.credentials().thenCompose(
             opt -> opt.<CompletionStage<Authentication>>map(
-                creds -> creds.yaml().thenApply(AuthFromYaml::new)
+                creds -> ((Credentials.FromStorageYaml) creds).yaml().thenApply(AuthFromYaml::new)
```
olegmoz, line 89: @olenagerasimova casting `creds` to `Credentials.FromStorageYaml` might cause problems in the future due to types mismatches in runtime. I have two solutions in mind:
- Signature of `CompletionStage<Optional<Credentials>> credentials()` method may be changed to `CompletionStage<Optional<Credentials.FromStorageYaml>> credentials()` in this class. `Settings` interface remains untouched.
- `CompletionStage<Authentication> auth()` method may be added to `Credentials` interface and creating `AuthFromYaml` is moved to `auth()` implementation in `Credentials.FromStorageYaml`.
What do you think?

olegmoz (changes requested):
@olenagerasimova thanks for the changes, but I have one concern regarding class casting. Please see the comment

src/main/java/com/artipie/YamlSettings.java
olenagerasimova: @olegmoz hm, I like the second option better, it will be more flexible and handy when we will have other `Authentication` and `Credentials` implementations. And technically we have credentials from env now, we can create corresponding `Credentials` implementation letter. 
Do you mind if I todo your suggestion?

olenagerasimova:
@olegmoz thanks, added todo

src/main/java/com/artipie/YamlSettings.java
olegmoz: @olenagerasimova yes, adding todo will be fine. Please mention removing this dynamic cast in the puzzle.

olegmoz (approved):
@olenagerasimova thanks, approved

0crat:
Job `gh:artipie/artipie#467` is not assigned, can't get performer

PR #476, opened 2020-08-20
#444 - GetUsersSlice introduced

olenagerasimova:
For  #444 implemented `GetUsersSlice` to obtain users list with urls.

### DIFF
--- a/src/main/java/com/artipie/api/artifactory/GetUsersSlice.java
+++ b/src/main/java/com/artipie/api/artifactory/GetUsersSlice.java
@@ -1,0 +1,110 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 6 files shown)


src/main/java/com/artipie/api/artifactory/GetUsersSlice.java
```diff
@@ -0,0 +1,110 @@
...
+                                );
+                                return new RsJson(json);
+                            }
+                        )
+                    ).orElse(CompletableFuture.completedFuture(StandardRs.NOT_FOUND))
+                )
+            );
+        } catch (final IOException err) {
+            Logger.error(this, err.getMessage());
+            res = new RsWithStatus(RsStatus.INTERNAL_ERROR);
```
olegmoz, line 92: @olenagerasimova I'd suggest to not translate error to response here, as it is done on upper level too. You could rethrow `IOException` wrapped into runtime exception or return `Response` which returns failed future on `send(...)` call

src/main/java/com/artipie/api/artifactory/GetUsersSlice.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        return res;
+    }
+
+    /**
+     * Returns json for user.
+     * @param name Username
+     * @param base Base url
```
olegmoz, line 100: @olenagerasimova parameters order got me confused, could you please put `base` before `name`?

olenagerasimova:
@olegmoz thanks, done

0crat:
Job `gh:artipie/artipie#476` is not assigned, can't get performer

PR #492, opened 2020-08-24
#444 - permissions API skeleton

olenagerasimova:
Part of #444 
Created skeleton to implement permissions API.

### DIFF
--- a/src/main/java/com/artipie/RepoPermissions.java
+++ b/src/main/java/com/artipie/RepoPermissions.java
@@ -100,18 +100,18 @@ public interface RepoPermissions {
 
         @Override
         public CompletionStage<Void> remove(final String repo) {
-            return null;
+            throw new UnsupportedOperationException("To be implemented");
         }
 
         @Override
         public CompletionStage<Void> addUpdate(final String repo, final String username,
             final String permission) {
-            return null;
+            throw new UnsupportedOperationException("Not implemented yet");
         }
 
         @Override
         public CompletionStage<Map<String, List<String>>> get(final String repo) {
-            return null;
+            throw new UnsupportedOperationException("To implemented later");
         }
 
         /**
--- a/src/main/java/com/artipie/RepoPermissions.java
+++ b/src/main/java/com/artipie/RepoPermissions.java
@@ -1,0 +1,130 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.artipie;
+
+import com.artipie.asto.Key;
+import com.artipie.asto.Storage;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
+
+/**
+ * Repository permissions settings.
+ * @since 0.10
+ */
+public interface RepoPermissions {
+
+    /**
+     * Artipie repositories list.
+     * @return Repository names list
+     */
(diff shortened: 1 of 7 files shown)


src/main/java/com/artipie/RepoPermissions.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
+
+/**
+ * Repository permissions settings.
+ * @since 0.10
+ */
+public interface RepoPermissions {
```
olegmoz, line 39: @olenagerasimova I'd suggest to extract individual repository permissions class from this one, so we can avoid complex methods like `addUpdate(String, String, String)` and return type like `CompletionStage<Map<String, List<String>>>`. What I suggest is something like this:
```java
interfaces RepoPermissionsRegistry {
  List<RepoPermissions> list();
  RepoPermissions get(String name);
  ...
}
interfaces RepoPermissions {
  void add(String username, String permission);
  ...
}
```
What do you think?

src/main/java/com/artipie/RepoPermissions.java
```diff
@@ -0,0 +1,130 @@
...
+        @Override
+        public CompletionStage<Void> addUpdate(final String repo, final String username,
+            final String permission) {
+            return null;
+        }
+
+        @Override
+        public CompletionStage<Map<String, List<String>>> get(final String repo) {
+            return null;
+        }
```
olegmoz, line 115: @olenagerasimova if I understand correctly, these methods need to be implemented. Am I right? It would be way more obvious to throw `UnsupporetedOperationException`, as it is common contract for not yet implemented methods. `null` value may look like a normal result and be propagated too far breaking the fail-fast principle.

Same goes for not implemented slices.

src/main/java/com/artipie/api/artifactory/AddUpdatePermissionSlice.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Artifactory `GET /api/security/permissions/{target}` endpoint, returns json with
+ * permissions (= repository) information.
+ * @since 0.10
+ * @todo #444:30min Implement GetPermissionSlice return repository permissions.
+ *  First, implement and test RepoPermissions.FromSettings#addUpdate(java.lang.String), then use
+ *  this method in this slice. Request json format can be found
+ *  https://www.jfrog.com/confluence/display/rtf/artifactory+rest+api, we should obtain information
+ *  from `repo.actions.users` fields.
```
olegmoz, line 40: @olenagerasimova the comment looks wrong, please update

src/test/java/com/artipie/RepoPermissionsFromSettingsTest.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+class RepoPermissionsFromSettingsTest {
+
+    @Test
+    void returnsRepoList() {
+        final Storage storage = new InMemoryStorage();
+        storage.save(new Key.From("one.yaml"), new Content.From(new byte[]{})).join();
+        storage.save(new Key.From("two.yaml"), new Content.From(new byte[]{})).join();
+        storage.save(new Key.From("abc"), new Content.From(new byte[]{})).join();
+        storage.save(new Key.From("three.yaml"), new Content.From(new byte[]{})).join();
```
olegmoz, line 46: @olenagerasimova `Content.EMPTY` was added in ASTO recently, it would make code less verbose if used here

olegmoz (changes requested):
@olenagerasimova thanks, please see the comments

src/main/java/com/artipie/RepoPermissions.java
olenagerasimova: @olegmoz hm, I do not see how one more interface will allow us to avoid returning `Map<String, List<String>>` for repo permissions, as we still need map `user -> permissions list`... The only thing we will avoid here - passings parameter `repo`

olenagerasimova:
@olegmoz thanks, I've corrected and answered, not sure about your suggestion about another interface...

src/main/java/com/artipie/RepoPermissions.java
olegmoz: @olenagerasimova basically we replace `Map<String, List<String>>` with `RepoPermissions`. `RepoPermissions` may implement `Map<...>` or have similar methods like `Set<String> users()` instead of `Set<String> keySet()` and `Set<String> permissions(String user)` instead of `List<String> get(String value)`.

It is all about making code more understandable. It's hard to understand what is what in type like `Map<String, List<String>>`.

src/main/java/com/artipie/api/artifactory/AddUpdatePermissionSlice.java
```diff
@@ -33,8 +33,8 @@
  * Artifactory `GET /api/security/permissions/{target}` endpoint, returns json with
```
olegmoz, line 33: @olenagerasimova `GET` method here looks wrong

olegmoz (changes requested):
@olenagerasimova thanks for the update. Answered and one minor issue found in latest commit

src/main/java/com/artipie/RepoPermissions.java
olenagerasimova: @olegmoz `RepoPermissions` still will have to return users and permissions for repository (map `username -> permissions list`) as `GetPermissionSlice` have to return following json format
```json
"users" : {
      "bob": ["r","w","m"],
      "alice" : ["d","w","n", "r"]
}
```
`Map<String, List<String>>` seems to be very convenient structure to build this json. Basically, I do not mind creating the interface you are suggesting, I just do not think it's necessary in this case, we can simply rename
`CompletionStage<Map<String, List<String>>> get(String repo)` to `CompletionStage<Map<String, List<String>>> permissions(String repo)` and extend javadoc.
Furthermore, having separate methods to obtain users and permissions from settings can lead to inconsistency as they can be changed while we are building json response in `GetPermissionSlice`.

olenagerasimova:
@olegmoz I've corrected javadoc and merged master

src/main/java/com/artipie/RepoPermissions.java
olegmoz: @olenagerasimova that makes sense, thanks

olegmoz (approved):
@olenagerasimova thanks, merging the PR

0crat:
Job `gh:artipie/artipie#492` is not assigned, can't get performer

Issue #464, opened 2020-08-20
ArtipieApi.java:68-70: Constructor decomposition This...


0pdd:
The puzzle `444-d6b8b13e` from #444 has to be resolved: 

https://github.com/artipie/artipie/blob/f832df1ed70a9b0b34efb48af8de734bc56a9340/src/main/java/com/artipie/api/ArtipieApi.java#L68-L70

The puzzle was created by @olenagerasimova on 19-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `444-d6b8b13e` has disappeared from the source code, that's why I closed this issue.

PR #505, opened 2020-08-25
#495 - GetPermissionSlice implemented

olenagerasimova:
Implemented `GetPermissionSlice` for #495

### DIFF
--- a/src/main/java/com/artipie/api/ArtipieApi.java
+++ b/src/main/java/com/artipie/api/ArtipieApi.java
@@ -29,6 +29,7 @@ import com.artipie.YamlPermissions;
 import com.artipie.api.artifactory.AddUpdateUserSlice;
 import com.artipie.api.artifactory.CreateRepoSlice;
 import com.artipie.api.artifactory.DeleteUserSlice;
+import com.artipie.api.artifactory.GetPermissionSlice;
 import com.artipie.api.artifactory.GetPermissionsSlice;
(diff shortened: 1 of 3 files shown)


src/main/java/com/artipie/api/ArtipieApi.java
```diff
@@ -62,6 +63,7 @@
  * Artipie API endpoints.
  * @since 0.6
  * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ * @checkstyle ClassFanOutComplexityCheck (500 lines)
```
olegmoz, line 66: @olenagerasimova I'd suggest to write a puzzle about reducing this class complexity

olegmoz (changes requested):
@olenagerasimova thanks, all is good except one thing, please see the comment

src/main/java/com/artipie/api/ArtipieApi.java
olenagerasimova: @olegmoz this class already has a puzzle to decompose constructor https://github.com/artipie/artipie/issues/464, I think that's enough considering the fact that it has no methods, ctor only.

olegmoz: @olenagerasimova ah, I see. That's clearly enough, thanks

olegmoz (approved):

0crat:
Job `gh:artipie/artipie#505` is not assigned, can't get performer

Issue #497, opened 2020-08-25
DeletePermissionSlice.java:36-40: Implement...


0pdd:
The puzzle `444-61eb484a` from #444 has to be resolved: 

https://github.com/artipie/artipie/blob/6abf31a5a47222f022d171f53e864b847853657b/src/main/java/com/artipie/api/artifactory/DeletePermissionSlice.java#L36-L40

The puzzle was created by @olenagerasimova on 25-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `444-61eb484a` has disappeared from the source code, that's why I closed this issue.

0crat:
Job was finished in 5 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

PR #506, opened 2020-08-26
#494 - RepoPermissions.FromSettings fully implemented

olenagerasimova:
Part of #494 and #497 
Artifactory method `PUT /api/security/permissions/{target}` request body json contains users and permissions in the following [format](https://www.jfrog.com/confluence/display/JFROG/Security+Configuration+JSON)
```
...
"users" : {
  "bob": ["read","write","manage"],
  "alice" : ["write","annotate", "read"]
}
...
```
So, I've
- introduced `UserPermission` class and changed `RepoPermissions#addUpdate` and `RepoPermissions#permissions` signature
- implemented and tested `RepoPermissions#addUpdate`
- implemented `RepoPermissions#remove` as it's part of `RepoPermissions#addUpdate`

### DIFF
--- a/src/main/java/com/artipie/RepoPermissions.java
+++ b/src/main/java/com/artipie/RepoPermissions.java
@@ -37,6 +37,7 @@ import hu.akarnokd.rxjava2.interop.SingleInterop;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -70,14 +71,14 @@ public interface RepoPermissions {
      * @param permissions Permissions list
      * @return Completion action
      */
-    CompletionStage<Void> addUpdate(String repo, List<UserPermission> permissions);
+    CompletionStage<Void> addUpdate(String repo, Collection<UserPermission> permissions);
 
     /**
      * Get repository permissions settings, returns users permissions list.
      * @param repo Repository name
      * @return Completion action with map with users and permissions
      */
-    CompletionStage<List<UserPermission>> permissions(String repo);
+    CompletionStage<Collection<UserPermission>> permissions(String repo);
 
     /**
      * {@link RepoPermissions} from Artipie settings.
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/RepoPermissions.java
```diff
@@ -61,25 +67,34 @@
     /**
      * Adds or updates repository permission.
      * @param repo Repository name
-     * @param username Username
-     * @param permission Permission name
+     * @param permissions Permissions list
      * @return Completion action
      */
-    CompletionStage<Void> addUpdate(String repo, String username, String permission);
+    CompletionStage<Void> addUpdate(String repo, List<UserPermission> permissions);
```
olegmoz, line 73: @olenagerasimova I am concerned about `List<UserPermission>`. It makes one think that users are ordered and order does matter. I do not think that this is the case and would suggest to change it to `Collection`. Same goes for `permissions` method return value

src/test/java/com/artipie/RepoPermissionsFromSettingsTest.java
```diff
@@ -103,6 +110,88 @@ void returnsEmptyMapWhenPermissionsAreNotSet() {
         );
     }
 
+    @Test
+    void updatesUserPermissions() throws IOException {
```
olegmoz, line 114: @olenagerasimova I think we miss the case when user has some permissions and we update the user by adding more permissions or removing some permissions

src/test/java/com/artipie/RepoPermissionsFromSettingsTest.java
```diff
@@ -103,6 +110,88 @@ void returnsEmptyMapWhenPermissionsAreNotSet() {
...
+            )
+            .toCompletableFuture().join();
+        MatcherAssert.assertThat(
+            this.permissionsForUser(repo, ann),
+            Matchers.contains(download)
+        );
+    }
+
+    @Test
+    void deletesPermissionSection() throws IOException {
```
olegmoz, line 165: @olenagerasimova this test will pass if `remove` method just clears all repositories. May we check that other repos stay intact after removing a repo?

olegmoz (changes requested):
@olenagerasimova thanks, please see the comments

olenagerasimova:
@olegmoz thanks, all fixed

olegmoz (approved):

0crat:
Job `gh:artipie/artipie#506` is not assigned, can't get performer

Issue #521, opened 2020-08-27
Docker repositories do not work with authentication


olegmoz:
When Docker client tries to perform an operation it first sends `GET /v2/` request. Response to this request should contain version header and `WWW-Authenticate` header if needed. Now Artipie always sends empty `200 OK` response for this endpoint. This behavior is implemented in `DockerRoutingSlice`. It works well in scenarios when authentication is not required. But when auth is required for other endpoints client does not receive `WWW-Authenticate` and does not tries to authenticate for any other requests. So in practice authenticated access does not work for Docker repositories.

0crat:
You @olegmoz[/z](https://www.0crat.com/u/olegmoz) need to have one of these roles in order to do "Request order start": `ARC`, `PO`; I'm sorry to say this, but at the moment you've got no roles in this project

PR #523, opened 2020-08-27
#499 - Added Docker local repository auth tests

### LINKED ISSUE #499: Remove "layout: flat" from proofs and examples
Layout should be made optional for configuration using `flat` value by default. With this done `layout: flat` could be removed from all proofs and examples

olegmoz:
Closes #499 
Added test cases for Docker repository when auth is required.
Th tests are disabled, because feature does not work at the moment. See #521


olenagerasimova (approved):
@olegmoz thanks, looks good to merge

Issue #470, opened 2020-08-20
ArtipieServer.java:83-89: Extract class for building...


0pdd:
The puzzle `449-3e9e55ee` from #449 has to be resolved: 

https://github.com/artipie/artipie/blob/5966489d2a1ed5829396e4e9d85d0370bb9d9c94/src/test/java/com/artipie/ArtipieServer.java#L83-L89

The puzzle was created by Oleg Mozzhechkov on 20-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #525, opened 2020-08-27
#509 - Extract generating repo permissions

### LINKED ISSUE #509: GetPermissionSliceTest.java:55-58: Create class to...
The puzzle `495-a27c3b37` from #495 has to be resolved: 

https://github.com/artipie/artipie/blob/edebc7f1eefaa3068b72ea7a635a411027bd7540/src/test/java/com/artipie/api/artifactory/GetPermissionSliceTest.java#L55-L58

The puzzle was created by olenagerasimova on 25-Aug-20. 

genryxy:
Closes #509 
Extract generating repo permissions from `GetPermissionSliceTest`, `AddUpdatePermissionSliceTest` and `RepoPermissionsFromSettingsTest` into class `BuildingRepoPermissions`.


src/test/java/com/artipie/BuildingRepoPermissions.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Class for generating repo permissions.
+ * @since 0.10
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+public final class BuildingRepoPermissions {
```
olenagerasimova, line 46: @genryxy I'd suggest to rename this class into `RepoPermissions` without `Building`

src/test/java/com/artipie/BuildingRepoPermissions.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Storage.
+     */
+    private final Storage storage;
```
olenagerasimova, line 50: @genryxy as this class main purpose is to build repo permissions, I think that it should have these permissions as a field. Also, sometimes adding this repo setting yaml to storage is not required

olenagerasimova (changes requested):
@genryxy please, check https://github.com/artipie/artipie/issues/470 - we have issuie to create whole config builder, I think that this class should be responsible for adding permission section, that why  I'd suggest to change this class conception:
- First, add users/permissions as a field and remove field for storage. For users/permissions we can use `RepoPermissions.UserPermission` class.
- Second, as permissions are part of the artipie repo configurations, let's create method to accept `YamlMapping` (yaml `repo` section) configuration and return this mapping with added permissions section. Thus, we will be able to use this class for building whole test config file.
- Third, other methods: let's create methods without `YamlMapping` parameter to build some default config with given permissions and methods with `Storage` and `Key` parameters to put config to storage, etc (what methods are necessary now). 
- Also, we can add empty ctor to build config without `permissions` section.

What do you think?

genryxy:
@olenagerasimova these changes will be useful and simplify using permissions, I will try to implement the proposed class concept

src/test/java/com/artipie/BuildingRepoPermissions.java
genryxy: @olenagerasimova Maybe, it would be better to rename this class into `UserRepoPermissions` because we have `RepoPermissions` interface with the class `UserPermission` inside that doesn't implement this interface. We need `RepoPermissions.Userpermission` field in the new class.

genryxy:
@olenagerasimova please take a look

src/test/java/com/artipie/RepoPerms.java
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * Add permissions to the repo section.
+     * @param reposection YamlMapping with repo section
+     * @return YamlMapping with repo section and permissions.
+     */
+    public YamlMapping addPermissions(final YamlMapping reposection) {
+        return RepoPerms.repoSectionAsBuilder(reposection)
+            .add(RepoPerms.PERMISSIONS, this.permsYaml())
+            .build();
```
olenagerasimova, line 105: @genryxy I think we should support `yaml` style here and modify given `reposection` by adding permissions section and return this modified yaml without building

src/test/java/com/artipie/RepoPerms.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Empty repo section with default values.
+     * @return Empty repo section.
+     */
+    static YamlMappingBuilder emptyRepoSection() {
```
olenagerasimova, line 112: @genryxy as this method is never used, I'd suggest to remove it

src/test/java/com/artipie/RepoPerms.java
```diff
@@ -0,0 +1,148 @@
...
+    public void saveSettings(final Storage storage, final String repo) {
+        storage.save(
+            new Key.From(String.format("%s.yaml", repo)),
+            new Content.From(
+                Yaml.createYamlMappingBuilder().add(
+                    "repo",
+                    Yaml.createYamlMappingBuilder()
+                        .add("type", "any")
+                        .add(RepoPerms.PERMISSIONS, this.permsYaml()).build()
+                ).build().toString().getBytes(StandardCharsets.UTF_8)
```
olenagerasimova, line 92: @genryxy I think we should call `addPermissions()` to create yaml file

olenagerasimova (changes requested):
@genryxy added several comments, take a look please

genryxy:
@olenagerasimova thanks, please take a look

olenagerasimova (approved):
@genryxy thanks, looks good to me

0crat:
Job `gh:artipie/artipie#525` is not assigned, can't get performer

Issue #545, opened 2020-09-02
List files/folders API endpoint


olegmoz:
We need to implement API for reading files metadata. There is such API endpoint in Artifactory API: https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API#ArtifactoryRESTAPI-FileList

We need it to list folders, and for the start let's assume that listing should not be recursive (e.g. one level of depth)

g4s8:
@olegmoz I'll copy relevant API spec here from Artifactory docs:

**Description**: Get a flat (the default) or deep listing of the files and folders (not included by default) within a folder.
For deep listing you can specify an optional depth to limit the results.
Optionally include a map of metadata timestamp values as part of the result (only properties are displayed in since 3.0.0).
folder inclusion since 2.3.2; checksum inclusion since: 2.3.3; include folder root path since: 2.5.2. Supported by all types of repositories.
**Usage**: `GET /api/storage/{repoKey}/{folder-path}?list[&deep=0/1][&depth=n][&listFolders=0/1][&mdTimestamps=0/1][&includeRootPath=0/1]`
 - `deep` - if `1`, then respect `depth` query param, `0` by default
 - `depth` - `hierarchy depth, use only if `deep` query param is `1`
 - `listFolders` - include folders if `1`, `0` by default
 - `mdTimestamps` - include timestampts
 - `includeRootPath` - include root path item `/`

**Produces**: `application/json (application/vnd.org.jfrog.artifactory.storage.FileList+json)`

**Sample Output**:
`GET /api/storage/libs-release-local/org/acme?list&deep=1&listFolders=1&mdTimestamps=1`
```json
{
"uri": "http://localhost:8081/artifactory/api/storage/libs-release-local/org/acme",
"created": "ISO8601",
"files" : [
  {
  "uri": "/archived",
  "size": "-1",
  "lastModified": "ISO8601",
  "folder": "true"
  },
  {
  "uri": "/doc.txt",
  "size": "253207",
  "lastModified": "ISO8601",
  "folder": "false",
  "sha1": "sha1Checksum",
  "mdTimestamps": { "properties" : "lastModified (ISO8601)" }
  },
  {
  "uri": "/archived/doc1.txt",
  "size": "253100",
  "lastModified": "ISO8601",
  "folder": "false",
  "sha1": "sha1Checksum",
  "mdTimestamps": { "properties" : "lastModified (ISO8601)" }
  }
]
}
```

0crat:
[...]

PR #548, opened 2020-09-03
STtorage key list formatting and Artifactory storage API

g4s8:
#545 - initial commit for Artifactory storage list API: added storage key formatting for JSON and other, added primitive `Slice` implementation.

### DIFF
--- a/src/main/java/com/artipie/api/artifactory/KeyList.java
+++ b/src/main/java/com/artipie/api/artifactory/KeyList.java
@@ -1,0 +1,97 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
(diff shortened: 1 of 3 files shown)


src/main/java/com/artipie/api/artifactory/KeyList.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * This class can print storage key items in different formats,
+ * e.g. for Artifactory API and as list HTML page, etc.
+ * @since 0.10
+ */
+public final class KeyList {
```
olenagerasimova, line 39: @g4s8 hm, I do no think this class should be introduced in `artipie/artipie`, it seems more like part of `SliceList` we never implemented in `http` (but we have issue https://github.com/artipie/http/issues/158 for that): this `SliceList` could except storage, base key and interface to format each key. What do you think?

olenagerasimova (changes requested):
@g4s8 one comment, please, let me know what do you think

src/main/java/com/artipie/api/artifactory/KeyList.java
g4s8: @olenagerasimova yes, it's a good point. I think it can be moved to http as a part of implementation of artipie/http#158 ticket

g4s8:
@olenagerasimova can we merge this and move this code into http as a part of https://github.com/artipie/http/issues/158 ticket?

olenagerasimova:
@g4s8 sure, thanks

olenagerasimova (approved):

0crat:
Job `gh:artipie/artipie#548` is not assigned, can't get performer

PR #549, opened 2020-09-04
#544 - Optional permissions in RepoConfig

olegmoz:
Part of #544 
Implementing global permissions support requires special handling of absence of permissions configured in repository. This requires to return `Optional<Permissions>` in `RepoConfig.permissions()` method. Logic for extracting permissions part from repo config moved from `YamlPermissions` to `RepoConfig` for that

PR #564, opened 2020-09-10
#549 - Added GlobalPermissions interface without implementation

olegmoz:
Part of #549 
Added `GlobalPermissions` interface without implementation

### DIFF
--- a/src/main/java/com/artipie/Settings.java
+++ b/src/main/java/com/artipie/Settings.java
@@ -27,8 +27,10 @@ import com.amihaiemil.eoyaml.Yaml;
 import com.amihaiemil.eoyaml.YamlMapping;
 import com.artipie.asto.Storage;
 import com.artipie.asto.memory.InMemoryStorage;
+import com.artipie.auth.GlobalPermissions;
 import com.artipie.http.auth.Authentication;
 import java.io.IOException;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/artipie/Settings.java
```diff
@@ -74,6 +76,13 @@
      */
     CompletionStage<Credentials> credentials();
 
+    /**
+     * Artipie global permissions.
+     *
+     * @return Global permissions if configured, empty - otherwise.
+     */
+    CompletionStage<Optional<GlobalPermissions>> permissions();
```
olenagerasimova, line 84: @olegmoz it seems to me that using `Optional` as a return type always adds some unnecessary complications to the code. As far as I understand the idea, absence of global permissions means that we simply use repo permissions as we do now. May be for this case we can simply create and use some special implementation (which simply returns permissions for repository) of `GlobalPermissions` and avoid returning `Optional` here?

olenagerasimova (changes requested):
@olegmoz one suggestion, please, take a look

olenagerasimova (changes requested):
@olegmoz one suggestion, take a look please

src/main/java/com/artipie/Settings.java
olegmoz: @olenagerasimova I agree that it adds additional complexity, but we need special handling of `GlobalPermissions` absence. See the table in comments of https://github.com/artipie/artipie/pull/549

0crat:
Job `gh:artipie/artipie#564` is not assigned, can't get performer

0crat:
Pull request [#564](https://github.com/artipie/artipie/issues/564) was not merged, no payment for `ARC`, see [§28](http://www.zerocracy.com/policy.html#28)

PR #566, opened 2020-09-10
#565 - Artifactory API corrections

### LINKED ISSUE #565: Artifactory permissions API: return 404 when repo does not exist
`DeletePermissionSlice` and `GetPermissionSlice` should return 404 if requested permissions target (repository) does not exist. Such behavior is not directly documented by artifactory, but I think it would be more correct than return 500 as we do now.
Also let's correct `AddUpdatePermissionSlice` javadoc.

olenagerasimova:
Closes #565 
Fixed `DeletePermissionSlice` and `GetPermissionSlice` to return 404 if repository does not exist.

### DIFF
--- a/src/main/java/com/artipie/api/artifactory/GetPermissionSlice.java
+++ b/src/main/java/com/artipie/api/artifactory/GetPermissionSlice.java
@@ -26,19 +26,24 @@ package com.artipie.api.artifactory;
 import com.artipie.RepoPermissions;
 import com.artipie.Settings;
 import com.artipie.api.RsJson;
+import com.artipie.asto.Key;
 import com.artipie.http.Response;
 import com.artipie.http.Slice;
 import com.artipie.http.async.AsyncResponse;
 import com.artipie.http.rs.RsStatus;
 import com.artipie.http.rs.RsWithStatus;
+import com.artipie.http.rs.StandardRs;
 import java.nio.ByteBuffer;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import javax.json.Json;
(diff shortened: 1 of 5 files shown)


src/test/java/com/artipie/api/artifactory/GetPermissionSliceTest.java
```diff
@@ -72,6 +72,18 @@ void returnsBadRequestOnInvalidRequest() {
         );
     }
 
+    @Test
+    void returnsNotFoundIfRepoDoesNotExists() {
+        final String repo = "pypi";
```
olegmoz, line 77: @olenagerasimova this variable is useless, we may just put into string

src/main/java/com/artipie/api/artifactory/GetPermissionSlice.java
```diff
@@ -65,12 +70,27 @@ public GetPermissionSlice(final Settings settings) {
...
+                                        )
+                                    );
+                            } else {
+                                res = CompletableFuture.completedStage(StandardRs.NOT_FOUND);
+                            }
+                            return res;
+                        }
+                    )
+            ).orElse(CompletableFuture.completedFuture(new RsWithStatus(RsStatus.BAD_REQUEST)))
+        );
```
olegmoz, line 93: @olenagerasimova we may wrap only the code in `opt.map(...)` with `AsyncResponse` and also return `RsWithStatus` without wrapping it into `CompletableFuture` in `.orElse(...)`. This will simplify code a little bit

olegmoz (changes requested):

olenagerasimova:
@olegmoz thanks, corrected

olegmoz (approved):
@olenagerasimova thanks!

0crat:
Job `gh:artipie/artipie#566` is not assigned, can't get performer

PR #577, opened 2020-09-11
#499: "flat" set as default layout

IlyaMoskva:
#499 Fixed and Reworked.


src/test/java/com/artipie/YamlSettingsTest.java
```diff
@@ -46,12 +46,27 @@
...
 
+    @Test
+    public void shouldSetFlatAsDefaultLayout() throws Exception {
+        final YamlSettings settings = new YamlSettings(
+            String.join(
+                "",
+                "meta:\n",
+                "  storage:\n",
+                "    type: s3\n",
+                "    path: /var/artipie/configs\n"
```
olegmoz, line 61: @IlyaMoskva do we really need to provide all that content to `YamlSettings`? I'd suggest to remove not required lines from the test to make it more focused and clear

src/test/java/com/artipie/YamlSettingsTest.java
```diff
@@ -46,12 +46,27 @@
...
+            )
+        );
+        MatcherAssert.assertThat(
+            settings.layout(),
+            Matchers.is("flat")
```
olegmoz, line 66: @IlyaMoskva using `new IsEqual<>()` here will be more in line with code style of the project, see https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md#code-style

olegmoz (changes requested):
@IlyaMoskva thanks, please see two comments

src/test/java/com/artipie/YamlSettingsTest.java
IlyaMoskva: Changed.

src/test/java/com/artipie/YamlSettingsTest.java
IlyaMoskva: Not all, you're right. Fixed.

src/test/java/com/artipie/YamlSettingsTest.java
```diff
@@ -56,14 +57,12 @@ public void shouldSetFlatAsDefaultLayout() throws Exception {
             String.join(
                 "",
                 "meta:\n",
-                "  storage:\n",
-                "    type: s3\n",
-                "    path: /var/artipie/configs\n"
+                "  storage:\n"
```
olegmoz, line 60: @IlyaMoskva and is `storage:` part really needed? We do not access `storage` in this test and I believe the string could be simplified to just `"meta:\n"

olegmoz (changes requested):
@IlyaMoskva thanks, please see one comment

src/test/java/com/artipie/YamlSettingsTest.java
IlyaMoskva: It breaks in constructor or in other place before. Just 'meta:' doesn't work for me.

olegmoz: @IlyaMoskva that's weird, there is just assignment in constructor. Could you please elaborate the problem?

IlyaMoskva: @olegmoz Exactly storage is not necessary here, but empty meta throws Exception in .yamlMapping(YamlSettings.KEY_META) call. It must contain at least something inside. I believe, it's out of scope for this ticket.

olegmoz: @IlyaMoskva got it, thanks

olegmoz (approved):

0crat:
Job `gh:artipie/artipie#577` is not assigned, can't get performer

Issue #569, opened 2020-09-11
Add emails to users


olegmoz:
Users in Artipie have only name and password at the moment. It would be nice to also have option to add email for each user. This email (if set) should be configurable via YAML and be supported by HTTP API as well

olenagerasimova:
@olegmoz do I understand correctly, that we are adding `email` field to `credentials.yaml` like this:
```yaml
credentials:
  alena:
    pass: plain:123
    email: alena@artipie.com
```

olegmoz:
@olenagerasimova yes, that's right. But `email` filed is optional and for now should be accessed only by HTTP API endpoints

olenagerasimova:
@olegmoz thanks, got it

Sammers21:
@olegmoz, how email can be useful?

olegmoz:
@Sammers21 this is for HTTP API implementation, see https://github.com/artipie/artipie/blob/master/ARTIFACTORY_API_SUPPORT.md#get-user-details

0pdd:
@olegmoz the puzzle [#589](https://github.com/artipie/artipie/issues/589) is still not solved.

0pdd:
@olegmoz the only puzzle [#589](https://github.com/artipie/artipie/issues/589) is solved here.

0pdd:
@olegmoz the puzzle [#589](https://github.com/artipie/artipie/issues/589) is still not solved.

PR #579, opened 2020-09-11
#569 - created Credentials.User

olenagerasimova:
Part of #569 
I've added `Credentials.User` class with username and email, changed methods of `Credentials` signature, implementations and usages. 
Next part would be to properly correct and extend tests.

### DIFF
--- a/src/main/java/com/artipie/Credentials.java
+++ b/src/main/java/com/artipie/Credentials.java
@@ -147,7 +147,7 @@ public interface Credentials {
                                 Optional.ofNullable(
                                     yaml.yamlMapping(FromStorageYaml.CREDENTIALS)
                                         .yamlMapping(name).string(FromStorageYaml.EMAIL)
-                                ).orElse(String.format("%s@artipie.com", name))
+                                )
                             );
                         }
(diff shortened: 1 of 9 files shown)


src/main/java/com/artipie/Credentials.java
```diff
@@ -274,4 +290,47 @@ public FromEnv(final Map<String, String> env) {
...
+
+        /**
+         * User name.
+         */
+        private final String uname;
+
+        /**
+         * User email.
+         */
+        private final String mail;
```
olegmoz, line 308: @olenagerasimova not every user has an email, so I think it will make more sense to make it `Optional`. If the field is required in certain `Slice` then we may put default value there, but not in all places where `User` is added/modified

olegmoz (changes requested):

olenagerasimova:
@olegmoz thanks, I'm not sure though that password should be part of the `Credentials.User` as it's not always required.

olegmoz (approved):

0crat:
Job `gh:artipie/artipie#579` is not assigned, can't get performer

Issue #535, opened 2020-08-31
Maven integration tests


g4s8:
Part of #389 Let's add an integration test similar to https://github.com/artipie/artipie/issues/449 to verify Maven repository:
 - Deploy and download release artifact
 - Deploy and download snapshot artifact
 - Download and cache artifact using remote proxy Maven slice

It can be implemented using `org.eclipse.aether` classes similar to https://github.com/artipie/maven-adapter/blob/master/src/test/java/com/artipie/maven/MavenHttpITCase.java

0crat:
@olegmoz[/z](https://www.0crat.com/u/olegmoz) the issue is closed not by @g4s8[/z](https://www.0crat.com/u/g4s8) (its creator); I won't close the order; please, re-open it and ask @g4s8[/z](https://www.0crat.com/u/g4s8) to close it

PR #580, opened 2020-09-11
#535 - Add mavenProxyIT

genryxy:
Part of #535 
Add `MavenProxyIT`. Artifact are downloading from `maven central`.

### DIFF
--- a/src/test/java/com/artipie/maven/MavenProxyIT.java
+++ b/src/test/java/com/artipie/maven/MavenProxyIT.java
@@ -25,7 +25,9 @@ package com.artipie.maven;
 
 import com.amihaiemil.eoyaml.Yaml;
 import com.artipie.ArtipieServer;
-import com.artipie.docker.junit.DockerClientSupport;
+import com.artipie.asto.Key;
+import com.artipie.asto.Storage;
+import com.artipie.asto.fs.FileStorage;
 import com.jcabi.log.Logger;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


olegmoz:
@olenagerasimova could you please review this PR?

src/test/java/com/artipie/maven/MavenProxyIT.java
```diff
@@ -0,0 +1,172 @@
...
+        this.cntn.close();
+    }
+
+    @Test
+    void shouldGetArtifactFromLocalAndSaveInCache() throws Exception {
+        final String artifact = "-Dartifact=aspectj:aspectj-ant:1.0.6:jar";
+        this.exec("mvn", "-s", "/home/settings.xml", "dependency:get", artifact);
+        MatcherAssert.assertThat(
+            this.exec(
+                "mvn", "-s", "/home/settings.xml", "dependency:get", artifact
```
olenagerasimova, line 115: @genryxy I see that this command is executed twice, is it really necessary?

olenagerasimova (changes requested):
@genryxy thanks, please, check my comment

src/test/java/com/artipie/maven/MavenProxyIT.java
genryxy: @olenagerasimova thanks. I try to verify that artifacts are cached. Maybe, it would be nice just to check that storage contains the key, but I have some troubles with it therefore implemented so. I'll try to check through storage and execute this command only once.

olenagerasimova: @genryxy have you tried simply check temp directory structure? You can add brake point, copy path and use `ls -la`. As far as see files should be `./subdir/repos/`. When you understand where files are, you can create `FileStorage` instance over subdir and check that files are loaded by corresponding keys

olenagerasimova (changes requested):

src/test/java/com/artipie/maven/MavenProxyIT.java
genryxy: @olenagerasimova yes, I tried to check temp directory. It has `artipie.yaml`, `settings`, `repos/my-maven.yaml`, `repos/_credentials.yml`

olegmoz:
@genryxy @olenagerasimova as I understand, we have Maven proxy broken at the moment. But may we merge the test in disabled state and enable it later when Maven proxy is fixed?

genryxy:
@olegmoz probably we can. In this case I'll need to commit changes with checking that file exist in storage after `dependency:get`.

olegmoz:
@olenagerasimova do you still some issues with this PR? to me it looks ready to be merged

olenagerasimova:
@olegmoz no, don't have any, please, merge

0crat:
Job `gh:artipie/artipie#580` is not assigned, can't get performer

PR #587, opened 2020-09-14
#548 - Tests for GetStorageSlice

olegmoz:
Part of #548 
Added tests for `GetStorageSlice`

### DIFF
--- a/src/test/java/com/artipie/api/artifactory/GetStorageSliceTest.java
+++ b/src/test/java/com/artipie/api/artifactory/GetStorageSliceTest.java
@@ -1,0 +1,132 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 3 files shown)


src/test/java/com/artipie/api/artifactory/GetStorageSliceTest.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    @TempDir
+    Path tmp;
+
+    @ParameterizedTest
+    @CsvSource({
+        "flat,my-lib",
+        "org,my-company/my-lib"
+    })
+    void shouldReturnExpectedData(final String layout, final String repo) {
```
olenagerasimova, line 77: @olegmoz `layout` is not used, maybe we can remove it and use `@ValueSource` here

olenagerasimova (changes requested):
@olegmoz just one comment, please, take a look

src/test/java/com/artipie/api/artifactory/GetStorageSliceTest.java
olegmoz: @olenagerasimova it is not used now, but it should be used later. `GetStorageSlice` accepts `Storage` as parameter now, but it should be changed to accept `Settings`, extract repo name from request path using `Settings.layout` and then take `Storage` by repo name. So I started off with wrtiting proper test and will change `GetStorageSlice` to do all that later

olegmoz:
@olenagerasimova answered your question, please review

olenagerasimova (approved):
@olegmoz thanks for the explanation, I see now

0crat:
Job `gh:artipie/artipie#587` is not assigned, can't get performer

Issue #574, opened 2020-09-11
Permission aliases in HTTP API


olegmoz:
In permissions API we need to support short aliases for permissions, `r` for read or download operations, `w` for write or upload operations

olenagerasimova:
@olegmoz as we agreed today the most suitable and universal way here is to support some kind of synonyms set for "read" and "write", so here is what I'm planning to do:
1. In http create enum with permissions `READ` and `WRITE` and varargs field for synonyms, create class to obtain enum instance by any string synonym 
2. Change `Permission.ByName` string `action` field to enum and add new ctor, use a set of synonyms to call `Permissions#allowed` with from the `Permission.ByName#allowed`
3. Change `AddUpdatePermissionSlice` to convert all possible sent permissions to `read` and `write` before saving them to config
4. Change `GetPermissionSlice` to convert all possible actions names from config (including asterisk as artifactory does not support it) to `read` and `write` before returning them

As far as I see first two steps should save us from changing anything in the adapters, I do not see any problems with docker either. 
Any thoughts here?

olegmoz:
@olenagerasimova design you propose sounds good, but I would not use enum for permission and create some interface instead. This will allow us to create custom permissions dynamically. For example we may need a read permissions for certain path, this cannot be done with enum.

0crat:
@olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) resigned from [#574](https://github.com/artipie/artipie/issues/574), since the job is not in scope anymore

PR #593, opened 2020-09-15
#574 - permissions mapping in AddUpdatePermissionSlice

olenagerasimova:
Part of #574 
I've added permissions mapping to `AddUpdatePermissionSlice` before saving them to repo config, possible permissions values are from [this](https://www.jfrog.com/confluence/display/JFROG/Security+Configuration+JSON#SecurityConfigurationJSON-application/vnd.org.jfrog.artifactory.security.PermissionTargetV2+json) docs.
Note that
- `annotate`, `managedXrayMeta` and `distribute` are not included as we do not know what they mean and how to translate them
- `admin` and `manage` are translated to `*` which means any actions will be allowed to these users

I'm planning to document it when we discuss, possibly change something and merge this PR.

### DIFF
--- a/src/main/java/com/artipie/api/artifactory/AddUpdatePermissionSlice.java
+++ b/src/main/java/com/artipie/api/artifactory/AddUpdatePermissionSlice.java
@@ -38,18 +38,38 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.json.Json;
 import javax.json.JsonObject;
-import javax.json.JsonValue;
+import org.cactoos.map.MapEntry;
+import org.cactoos.map.MapOf;
 import org.reactivestreams.Publisher;
 
 /**
  * Artifactory `PUT /api/security/permissions/{target}` endpoint, updates `permissions` section
  * in repository section.
  * @since 0.10
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  */
-@SuppressWarnings("PMD.AvoidDuplicateLiterals")
 public final class AddUpdatePermissionSlice implements Slice {
+
+    /**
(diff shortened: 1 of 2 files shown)


olegmoz:
@olenagerasimova I am not sure at this point that we need this translation when actions added in https://github.com/artipie/http/pull/239

Permissions may be stored as-is and translated when permission is checked. Saving permissions as-is will also allow to return expected values in `GET` requests.

olenagerasimova:
@olegmoz hm, as these are artifactory-like terms and permissions, I'd not want them to penetrate dipper than this API layer. 
I consider the whole permission-translation task consist of two parts:
1. Unify our internal terms (actions from https://github.com/artipie/http/pull/239 allow us to achieve this goal), actions names contains terms from our adapters only
2. Translate artifactory-like terms to artipie terms and back, this PR serves this goal partly

What do you say?

olegmoz:
@olenagerasimova I also would rather avoid passing Artifactory API logic beyond the slices, but I am not sure how permissions translated into `*` will work. First we receive request to grant permission `admin` to some user, we store `*` instead. How will translate `*` back to `admin` later? How will we know that it is `admin` permission and not `manage`? Can't think of a solution, do you have some solution in mind?

And for `r`, `read, `w`, `write` and others we have translations in `Action.ByString` already, so no need to do it here

olenagerasimova:
@olegmoz I understand that we will not be able to return exact excepted values, but is it really required? 
As for `*` I was thinking to translate it into `manage` as artifactory "PermissionTargetV2+json" does not have "admin".

If we are not going to merge it, then we should extend `Actions.Standard` with `admin` or `manage` value and probably add some values to existing items...

olegmoz (approved):
@olenagerasimova thanks, got it. As the translation is not ambiguous the solution is nice. And we won't have external permission names deep in our code

0crat:
Job `gh:artipie/artipie#593` is not assigned, can't get performer

PR #599, opened 2020-09-16
Repo on separeted port

### LINKED ISSUE #570: Repositories on own ports
Right now connections are accepted on single port and all repositories are available with separation by HTTP path prefix. It would be useful to allow configurations when Artipie listens on multiple ports and every port is for it's own repository.

That would be especially useful for Docker repositories, as their API is not tailored to allow multiple repositories on single port

Sammers21:
Closes #570

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -271,6 +271,11 @@ SOFTWARE.
       <version>4.2.0</version>
     </dependency>
     <dependency>
+      <groupId>org.javatuples</groupId>
+      <artifactId>javatuples</artifactId>
+      <version>1.2</version>
+    </dependency>
+    <dependency>
       <groupId>commons-cli</groupId>
       <artifactId>commons-cli</artifactId>
       <version>1.4</version>
--- a/src/main/java/com/artipie/RepoConfig.java
+++ b/src/main/java/com/artipie/RepoConfig.java
@@ -99,6 +99,14 @@ public final class RepoConfig {
     }
 
     /**
+     * Repo port.
+     * @return Repo port
+     */
+    public Integer port() {
+        return this.repoConfig().integer("port");
+    }
+
+    /**
      * Repository path.
      * @return Async string of path
      */
--- a/src/main/java/com/artipie/SliceFromConfig.java
+++ b/src/main/java/com/artipie/SliceFromConfig.java
@@ -139,7 +139,12 @@ public final class SliceFromConfig extends Slice.Wrap {
         final Permissions permissions = new LoggingPermissions(
             cfg.permissions().orElse(Permissions.FREE)
         );
-        final Pattern prefix = new PathPattern(settings).pattern();
+        final Pattern prefix;
+        if (cfg.port() == -1) {
+            prefix = new PathPattern(settings).pattern();
+        } else {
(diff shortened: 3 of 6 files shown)


pom.xml
```diff
@@ -270,6 +270,11 @@ SOFTWARE.
       <artifactId>handlebars</artifactId>
       <version>4.2.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.javatuples</groupId>
+      <artifactId>javatuples</artifactId>
+      <version>1.2</version>
+    </dependency>
```
olegmoz, line 277: @Sammers21 please avoid using this library. Reading code with tuples is hard. Creating custom class of finding other solution is preferred

src/main/java/com/artipie/RepoConfig.java
```diff
@@ -98,6 +98,14 @@ public String type() {
         return this.string("type");
     }
 
+    /**
+     * Repo port.
+     * @return Repo port
+     */
+    public Integer port() {
```
olegmoz, line 105: @Sammers21 I'd suggest to use `OptionalInt` here instead of `Integer`, to avoid nullability problems and represent absent `port` as empty value instead of `-1` as this YAML library does.

src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -139,7 +139,12 @@ static Slice build(final Settings settings, final Authentication auth,
         final Permissions permissions = new LoggingPermissions(
             cfg.permissions().orElse(Permissions.FREE)
         );
-        final Pattern prefix = new PathPattern(settings).pattern();
+        final Pattern prefix;
+        if (cfg.port() == -1) {
+            prefix = new PathPattern(settings).pattern();
+        } else {
+            prefix = Pattern.compile("()(/.*)?");
+        }
```
olegmoz, line 147: @Sammers21 I am not sure that it's okey to not serve only repositories without `port` on default port. Instead we should serve repositories with host on their host and on default host by name as well

src/main/java/com/artipie/VertxMain.java
```diff
@@ -152,6 +167,65 @@ public static void main(final String... args) throws IOException, ParseException
...
+                            "Slice at %s started at port:%d",
+                            key.string(),
+                            sport
+                        );
+                    } catch (final IOException exc) {
+                        throw new IllegalStateException(exc);
+                    }
+                }
+            );
+    }
```
olegmoz, line 227: @Sammers21 this method is to complex to understand and contains a lot of logic taken from other classes. Please extract this logic to separate class so it could be understood

olegmoz (changes requested):
@Sammers21 thanks, please see multiple comments. Also this PR cannot be accepted due to lack of tests. I'd suggest to break this PR into smaller ones with proper tests

0crat:
Job `gh:artipie/artipie#599` is not assigned, can't get performer

0crat:
Pull request [#599](https://github.com/artipie/artipie/issues/599) was not merged, no payment for `ARC`, see [§28](http://www.zerocracy.com/policy.html#28)

PR #619, opened 2020-09-22
#616 - Simplify user() method

### LINKED ISSUE #616: AuthFromYaml.java:45-47: Simplify `user()` method and get...
The puzzle `572-84159cfe` from #572 has to be resolved: 

https://github.com/artipie/artipie/blob/c61372a6934f8590e815758078e291e1667b3975/src/main/java/com/artipie/auth/AuthFromYaml.java#L45-L47

The puzzle was created by olenagerasimova on 21-Sep-20. 

genryxy:
Closes #616 
Simplify `user()` method and get rid of nested if expressions. Also get rid of `check(String, String)` method and use `check()` with three params instead.

### DIFF
--- a/src/main/java/com/artipie/auth/AuthFromYaml.java
+++ b/src/main/java/com/artipie/auth/AuthFromYaml.java
@@ -42,14 +42,8 @@ import org.apache.commons.codec.digest.DigestUtils;
  *  |   pass: "plain:123"
  *  This configuration would mean that Joe should use his email to login instead of username. Do not
  *  forget to validate credentials, logins should be unique.
- * @todo #572:30min Simplify `user()` method and get rid of nested if expressions: create separate
- *  method to obtain password type either from separate `type` yaml field or from `pass`, get rid of
- *  `check(String, String)` method and use `check()` with three params instead.
  */
-@SuppressWarnings({
-    "PMD.AvoidDeeplyNestedIfStmts",
-    "PMD.AvoidDuplicateLiterals",
-    "PMD.ConfusingTernary"})
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
 public final class AuthFromYaml implements Authentication {
 
     /**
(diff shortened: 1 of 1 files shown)


olegmoz:
@olenagerasimova could you please review this one?

src/main/java/com/artipie/auth/AuthFromYaml.java
```diff
@@ -74,24 +68,13 @@ public AuthFromYaml(final YamlMapping cred) {
...
-                        );
-                    }
-                } else {
-                    if (check(stored, pass)) {
-                        res = Optional.of(
-                            new User(user, AuthFromYaml.groups(users.yamlMapping(user)))
-                        );
-                    }
-                }
+            final String type = users.yamlMapping(user).string("type");
```
olenagerasimova, line 73: @genryxy passing nullable `type` into `check` method does not seem elegant, and also method with name `check` does not suppose to extract password type... So, I'd suggest create separate method to obtain password type (simply extract this logic from `check`) and then call `check` method with known password type.

olenagerasimova (changes requested):
@genryxy thanks, please, check my comment

src/main/java/com/artipie/auth/AuthFromYaml.java
genryxy: @olenagerasimova thanks. If I obtain password type from the method it becomes more difficult to distinguish password that have type at the beginning `plain:123` from the password `123`. I can pass `user` and `users` to `check()` method instead of nullable `type` but in this case method will have 4 parameters. Or would it be better to find a solution that doesn't pass a value to the method?

olenagerasimova: @genryxy I see, sorry for the misguided suggestion. Maybe we can make `type` `Optional<String>` to avoid null and rename `check` method properly? The name of the method should show shortly explain what the method does.

genryxy:
@olenagerasimova thanks, made `type` `Optional<String>`, renamed method `check`

olegmoz:
@genryxy be sure to re-request review from reviewer when you have everything is fixed. Otherwise the reviewer cannot see that PR is ready and requires attention

olenagerasimova (approved):
@genryxy thanks
@olegmoz seems good to me

0crat:
Job `gh:artipie/artipie#619` is not assigned, can't get performer

Issue #573, opened 2020-09-11
User groups support


olegmoz:
We should implement support for groups, so users may be assigned to multiple groups and some permissions may be granted to whole group and not individual users

One example is `readers` group, see https://github.com/artipie/artipie/issues/572

0crat:
You @olegmoz[/z](https://www.0crat.com/u/olegmoz) need to have one of these roles in order to do "Request order start": `ARC`, `PO`; I'm sorry to say this, but at the moment you've got no roles in this project

olenagerasimova:
@olegmoz hm, I think we can close this ticket as we support groups for users now and I continue my work on readers group support and part of #572

0crat:
Job `gh:artipie/artipie#573` is not assigned, can't get performer

Issue #572, opened 2020-09-11
Readers group support


olegmoz:
All users added using HTTP API should be automatically added to `readers` group and repositories created using HTTP API should be created with read permissions granted to `readers` group

olenagerasimova:
@olegmoz are we going to implement this by adding `read` permissions for `*` (everyone) in every repository added/updated with HTTP API?

olegmoz:
@olenagerasimova with this approach to revoke read access one will need to make changes in all repositories. That may do the trick as temporary solution, but I think we should actually implement groups. Groups going to be useful for big Artipie installations with many repositories and users

olenagerasimova:
@olegmoz as we discussed, here is my plan:
1. Prepare `http` by changing `Authentication` signature https://github.com/artipie/artipie/issues/573 
2. Release `http` and update it in every adapter (maybe Alexandr can help me here?)
3. Update `http` in artipie and change `AuthFromYaml` to return group list
4. Update rest api clices to add "reader" group to users and repo permissions

Sounds ok?

olenagerasimova:
@olegmoz let's discuss format for save user groups in `credentials.yaml`:
```yaml
credentials:
  artipie:
    pass: plain:artipie
    groups: 
      - reader
      - a-team
      - c-team
  john:
    pass: plain:123
    groups: 
      - reader
      - b-team
``` 
Is this format ok?

olegmoz:
@olenagerasimova yes, the format looks nice

0pdd:
@olegmoz the puzzle [#616](https://github.com/artipie/artipie/issues/616) is still not solved.

0pdd:
@olegmoz the only puzzle [#616](https://github.com/artipie/artipie/issues/616) is solved here.

PR #627, opened 2020-09-24
#572 - add reader group and permission

olenagerasimova:
Part of #573 
- user created from `AddUpdateUserSlice` always is in `readers` group
- repo permissions updated/created from `AddUpdatePermissionSlice` always include `read` right for `readers` groups

### DIFF
--- a/src/main/java/com/artipie/api/artifactory/AddUpdatePermissionSlice.java
+++ b/src/main/java/com/artipie/api/artifactory/AddUpdatePermissionSlice.java
@@ -45,6 +45,7 @@ import java.util.stream.Collectors;
 import javax.json.Json;
 import javax.json.JsonObject;
 import javax.json.JsonString;
+import org.cactoos.list.ListOf;
 import org.cactoos.map.MapEntry;
 import org.cactoos.map.MapOf;
 import org.reactivestreams.Publisher;
@@ -114,7 +115,7 @@ public final class AddUpdatePermissionSlice implements Slice {
     }
 
     /**
-     * Converts json permissions into list of {@link RepoPermissions.UserPermission}.
+     * Update repo permissions.
      * @param json Json body
      * @param name Repository name
      * @return True - if JSON is valid and update performed,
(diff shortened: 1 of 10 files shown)


src/main/java/com/artipie/api/artifactory/AddUpdatePermissionSlice.java
```diff
@@ -142,6 +143,7 @@ public Response response(final String line, final Iterable<Map.Entry<String, Str
                     )
                 )
         );
+        res.add(new RepoPermissions.PermissionItem("/readers", new ListOf<String>("read")));
```
olegmoz, line 146: @olenagerasimova `/readers` group should be added only if new repository is created. Otherwise it won't be possible to revoke permissions for the group

src/main/java/com/artipie/api/artifactory/AddUpdateUserSlice.java
```diff
@@ -108,11 +108,12 @@ public Response response(final String line, final Iterable<Map.Entry<String, Str
             json -> {
                 final Optional<Pair<Users.User, String>> res;
                 if (json.containsKey(pswd) && json.containsKey(email)) {
-                    List<String> list = Collections.emptyList();
+                    List<String> list = new ArrayList<>(1);
                     if (json.containsKey(groups)) {
                         list = json.getJsonArray(groups).getValuesAs(JsonString.class)
                             .stream().map(JsonString::getString).collect(Collectors.toList());
                     }
+                    list.add("readers");
```
olegmoz, line 116: @olenagerasimova same as with adding `readers` group to repository, this should be done only on new user creation

olegmoz (changes requested):

src/main/java/com/artipie/api/artifactory/AddUpdatePermissionSlice.java
olenagerasimova: @olegmoz are you sure such behavior will be correct from our colleagues point of view? As far as I understood, users always have `read` permission

olegmoz: @olenagerasimova I tried to lookup requirements about it and cannot really find direct answer. So let's assume `readers` group is always present in repository and all users are in `readers` group.

Still there is a problem in this line: we need to add `readers` group to the list only if it is absent in request. Otherwise it is going to be duplicated if user is loaded and the saved as-is. The same goes for adding `readers` group to user in `AddUpdateUserSlice` class.

olenagerasimova:
@olegmoz thanks, as for adding `readers` group - I've left the logic as is, also I've added `groups` field parsing in `AddUpdatePermissionSlice` which I missed yesterday

olegmoz (approved):
@olenagerasimova thanks for the fixes, merging

0crat:
Job `gh:artipie/artipie#627` is not assigned, can't get performer

PR #629, opened 2020-09-24
#570 - No repo name prefix on repository ports

### LINKED ISSUE #570: Repositories on own ports

olegmoz:
Closes #570 
No repo name prefix in path required when accessing repository on it's own port


olenagerasimova (approved):
@olegmoz thanks, looks good to merge

0crat:
Job `gh:artipie/artipie#629` is not assigned, can't get performer

Issue #603, opened 2020-09-17
Files repository integration tests


olegmoz:
Add integration tests that verify that Curl can upload and download binary files with authentication enabled and disabled

0crat:
You @olegmoz[/z](https://www.0crat.com/u/olegmoz) need to have one of these roles in order to do "Request order start": `ARC`, `PO`; I'm sorry to say this, but at the moment you've got no roles in this project

0crat:
Job `gh:artipie/artipie#603` is not assigned, can't get performer

olegmoz:
@genryxy it looks like we have all the tests needed: upload, download and auth coverage. I think we are ready to close the issue or we are missing something?

genryxy:
@olegmoz yeah, this question can be closed

PR #630, opened 2020-09-25
#603 - Check PUT and GET with auth disabled

genryxy:
Part of #603 
Add IT case for files repository. Verify `PUT` and `GET` with authentication disabled.

### DIFF
--- a/src/test/java/com/artipie/FilesRepoITCase.java
+++ b/src/test/java/com/artipie/FilesRepoITCase.java
@@ -33,6 +33,7 @@ import com.jcabi.log.Logger;
 import java.nio.file.Path;
 import java.util.regex.Pattern;
 import org.hamcrest.MatcherAssert;
+import org.hamcrest.core.IsEqual;
 import org.hamcrest.core.StringContains;
 import org.hamcrest.text.MatchesPattern;
 import org.junit.jupiter.api.AfterEach;
@@ -83,7 +84,7 @@ final class FilesRepoITCase {
     @BeforeEach
     void init() throws Exception {
         this.storage = new FileStorage(this.tmp);
-        this.server = new ArtipieServer(this.tmp, "my-file", this.configs());
+        this.server = new ArtipieServer(this.tmp, "my-file", this.config());
         this.port = this.server.start();
         this.server.start();
         Testcontainers.exposeHostPorts(this.port);
@@ -93,22 +94,15 @@ final class FilesRepoITCase {
             .withFileSystemBind(this.tmp.toString(), "/home");
         this.cntn.start();
         this.exec("yum", "-y", "install", "curl");
-        this.exec("yum", "-y", "install", "maven");
     }
 
     @Test
-    void downloadsArtifact() throws Exception {
-        final String url = "http://host.testcontainers.internal:%d/my-file/helloworld-src/pom.xml";
+    void curlGetShouldReceiveFile() throws Exception {
+        final String url = "http://host.testcontainers.internal:%d/my-file/file-repo/curl.txt";
         this.addFilesToStorage(
-            "helloworld-src", new Key.From("my-file", "helloworld-src")
+            "file-repo", new Key.From("repos", "my-file", "file-repo")
         );
         MatcherAssert.assertThat(
-            "curl PUT doesn't work properly",
-            this.exec("curl", "-i", "-X", "PUT", String.format(url, this.port)),
-            new StringContains("HTTP/1.1 201 Created")
-        );
-        MatcherAssert.assertThat(
-            "curl GET doesn't work properly",
             this.exec("curl", "-i", "-X", "GET", String.format(url, this.port)),
             new MatchesPattern(
                 Pattern.compile(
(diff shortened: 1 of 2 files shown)


src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -0,0 +1,163 @@
...
+        this.port = this.server.start();
+        this.server.start();
+        Testcontainers.exposeHostPorts(this.port);
+        this.cntn = new GenericContainer<>("centos:centos8")
+            .withCommand("tail", "-f", "/dev/null")
+            .withWorkingDirectory("/home/")
+            .withFileSystemBind(this.tmp.toString(), "/home");
+        this.cntn.start();
+        this.exec("yum", "-y", "install", "curl");
+        this.exec("yum", "-y", "install", "maven");
```
olegmoz, line 96: @genryxy it is not needed to install `maven` for these tests, we only need `curl`

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -0,0 +1,163 @@
...
+    void downloadsArtifact() throws Exception {
+        final String url = "http://host.testcontainers.internal:%d/my-file/helloworld-src/pom.xml";
+        this.addFilesToStorage(
+            "helloworld-src", new Key.From("my-file", "helloworld-src")
+        );
+        MatcherAssert.assertThat(
+            "curl PUT doesn't work properly",
+            this.exec("curl", "-i", "-X", "PUT", String.format(url, this.port)),
+            new StringContains("HTTP/1.1 201 Created")
+        );
```
olegmoz, line 109: @genryxy here we add the same file to storage twice: first with `addFilesToStorage` and then using `curl`. I'd say in test checking download we should add file to storage  using `addFilesToStorage` method and create another test that runs `curl PUT` and checks that file is present in `Storage` after that.

You may just remove call of `curl PUT` for now and add another test later that checks `curl PUT` runs normally and adds file to `Storage` at expected path

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -0,0 +1,163 @@
...
+        this.exec("yum", "-y", "install", "curl");
+        this.exec("yum", "-y", "install", "maven");
+    }
+
+    @Test
+    void downloadsArtifact() throws Exception {
+        final String url = "http://host.testcontainers.internal:%d/my-file/helloworld-src/pom.xml";
+        this.addFilesToStorage(
+            "helloworld-src", new Key.From("my-file", "helloworld-src")
+        );
```
olegmoz, line 104: @genryxy please do not use a `pom.xml` file for this test, it is misguiding. When one sees `pom.xml` it means we are talking about Maven. And Maven has nothing to do with it. Please just create a file with some text at some path, so it will not give wrong idea to reader.

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -0,0 +1,163 @@
...
+            "curl GET doesn't work properly",
+            this.exec("curl", "-i", "-X", "GET", String.format(url, this.port)),
+            new MatchesPattern(
+                Pattern.compile(
+                    // @checkstyle LineLengthCheck (1 line)
+                    "HTTP\\/1.1 200 OK[\\r\\n]{0,2}Content-Type: application\\/octet-stream[\\r\\n \\S]*"
+                )
+            )
+        );
+        this.exec("mvn", "clean");
```
olegmoz, line 120: @genryxy running `mvn clean` command here is not needed, please remove it

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -0,0 +1,163 @@
...
+        this.server.stop();
+        this.cntn.stop();
+    }
+
+    private String exec(final String... command) throws Exception {
+        Logger.debug(this, "Command:\n%s", String.join(" ", command));
+        return this.cntn.execInContainer(command).getStdout();
+    }
+
+    private String configs() {
```
olegmoz, line 134: @genryxy do not see the reason why name is plural, e.g. `configS`, why not just `config`?

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Test
+    void downloadsArtifact() throws Exception {
+        final String url = "http://host.testcontainers.internal:%d/my-file/helloworld-src/pom.xml";
+        this.addFilesToStorage(
+            "helloworld-src", new Key.From("my-file", "helloworld-src")
+        );
+        MatcherAssert.assertThat(
+            "curl PUT doesn't work properly",
```
olegmoz, line 106: @genryxy it's common to write what we expect, not what what happend on a failure. Meaning this would be better change to `curl PUT does work properly`. Same goes for other assert comment about `curl GET`

olegmoz (changes requested):
@genryxy thanks for the PR, please see some comments

src/test/java/com/artipie/FilesRepoITCase.java
genryxy: @olegmoz in some case I get the error as https://github.com/artipie/artipie/pull/537/checks?check_run_id=1081335703#step:5:1028 In the previous case running `mvn clean` command helped to fix. I can try to push without maven

olegmoz: @genryxy that was in Maven test, this test is absolutely unrelated

genryxy: @olegmoz understood, thanks

genryxy:

olegmoz (approved):
@g4s8 thanks, it's all good

genryxy:
@olegmoz I have `java.io.IOException: Failed to delete temp directory` I don't know exactly why it appears but `mvn clean` solves this problem

olegmoz:
> @olegmoz I have `java.io.IOException: Failed to delete temp directory` I don't know exactly why it appears but `mvn clean` solves this problem

I see that build failed in unrelated test:
```
Errors: 
2020-09-25T09:59:50.1115201Z [ERROR]   DockerLocalAuthIT.shouldFailPushIfNoWritePermission � IO Failed to delete temp...
2020-09-25T09:59:50.1116312Z [INFO] 
2020-09-25T09:59:50.1116715Z [ERROR] Tests run: 11, Failures: 0, Errors: 1, Skipped: 1
```
It seems to be random problem on host machine, going to try run it once again

0crat:
Job `gh:artipie/artipie#630` is not assigned, can't get performer

PR #639, opened 2020-09-25
#603 - Verify curl request with authentication 

### LINKED ISSUE #603: Files repository integration tests

genryxy:
Closes #603 
Verify `curl` `PUT` and `GET` with authentication enabled.

### DIFF
--- a/src/test/java/com/artipie/FilesRepoITCase.java
+++ b/src/test/java/com/artipie/FilesRepoITCase.java
@@ -24,6 +24,8 @@
 package com.artipie;
 
 import com.amihaiemil.eoyaml.Yaml;
+import com.amihaiemil.eoyaml.YamlMapping;
+import com.amihaiemil.eoyaml.YamlMappingBuilder;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
 import com.artipie.asto.blocking.BlockingStorage;
@@ -37,11 +39,12 @@ import org.hamcrest.core.IsEqual;
 import org.hamcrest.core.StringContains;
 import org.hamcrest.text.MatchesPattern;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledOnOs;
 import org.junit.jupiter.api.condition.OS;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.testcontainers.Testcontainers;
 import org.testcontainers.containers.GenericContainer;
 
@@ -50,9 +53,18 @@ import org.testcontainers.containers.GenericContainer;
  * @since 0.11
  * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  */
-@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+@SuppressWarnings({
+    "PMD.AvoidDuplicateLiterals",
+    "PMD.TooManyMethods"
+})
 @EnabledOnOs({OS.LINUX, OS.MAC})
 final class FilesRepoITCase {
+
+    /**
+     * Url for request.
+     */
+    private static final String URL =
+        "http://host.testcontainers.internal:%d/my-file/file-repo/curl.txt";
 
     /**
      * Temporary directory for all tests.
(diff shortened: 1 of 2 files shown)


src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -81,29 +93,18 @@
...
+        this.init(this.config(anonymous));
         this.addFilesToStorage(
             "file-repo", new Key.From("repos", "my-file", "file-repo")
         );
         MatcherAssert.assertThat(
-            this.exec("curl", "-i", "-X", "GET", String.format(url, this.port)),
+            this.exec(
+                "curl", "-i", "-X", "GET", String.format(FilesRepoITCase.URL, this.port),
+                this.flag(anonymous), this.user(anonymous, ArtipieServer.ALICE)
+            ),
```
olegmoz, line 107: @genryxy methods `flag` and `user` are tightly coupled, they are used together in same way with same argument all the time. I'd suggest to refactor `exec` method to accept `anonymous` flag and add necessary args to the command when enabled.

Maybe we should even rename `exec` to `curl` and omit first string argument, as it is always `"curl"` in this test class.

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -130,8 +135,35 @@ void curlPutShouldSaveFile() throws Exception {
...
+    void curlPutShouldFailWithForbidden() throws Exception {
+        this.init(this.config(false));
+        MatcherAssert.assertThat(
+            this.exec(
+                "curl", "-i", "-X", "PUT", String.format(FilesRepoITCase.URL, this.port),
+                this.flag(false), this.user(false, ArtipieServer.BOB)
+            ),
+            new StringContains("HTTP/1.1 403 Forbidden")
+        );
+    }
```
olegmoz, line 163: @genryxy please also add a test to see that `download` permission is checked as well. That will require to do `GET` with user who lacks `download` permission

olegmoz (changes requested):

genryxy:
@olegmoz thanks, refactored method `exec` and added test for `GET` with user who lacks `download` permission

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -168,9 +154,20 @@ void tearDown() {
         this.cntn.stop();
     }
 
-    private String exec(final String... command) throws Exception {
-        Logger.debug(this, "Command:\n%s", String.join(" ", command));
-        return this.cntn.execInContainer(command).getStdout();
+    private String curl(final boolean anonymous, final String action,
+        final ArtipieServer.User usr) throws Exception {
```
olegmoz, line 158: @genryxy I'd suggest to refactor the method so it accepts `Optional<ArtipieServer.User>` instead of ArtipieServer.User` and remove `anonymous` flag. This way we avoid input that makes no sense with `anonymous=true` and some user specified

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -168,9 +154,20 @@ void tearDown() {
...
+        final String url = "http://host.testcontainers.internal:%d/my-file/file-repo/curl.txt";
+        final List<String> cmdlst = new ArrayList<>(
+            Arrays.asList(
+                "curl", "-i", "-X", action, String.format(url, this.port)
+            )
+        );
+        cmdlst.add(this.flag(anonymous));
+        cmdlst.add(this.user(anonymous, usr));
```
olegmoz, line 166: @genryxy the code would be easier to understand if just inline these `flag` and `user` methods, the issue with these methods tightly coupled and called with same arguments in every place still stands

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -168,9 +154,20 @@ void tearDown() {
...
+        String[] cmdarr = new String[cmdlst.size()];
+        cmdarr = cmdlst.toArray(cmdarr);
```
olegmoz, line 168: @genryxy why not just change these two lines into `String[] cmdarr = cmdlst.toArray(new String[cmdlst.size()]);`, first assignment is just not needed and is mind-boggling

olegmoz (changes requested):
@genryxy thanks for the changes, there are still issues in the PR. Please see the comments

genryxy:

src/test/java/com/artipie/FilesRepoITCase.java
```diff
@@ -154,18 +158,18 @@ void tearDown() {
...
                 "curl", "-i", "-X", action, String.format(url, this.port)
             )
         );
-        cmdlst.add(this.flag(anonymous));
-        cmdlst.add(this.user(anonymous, usr));
-        String[] cmdarr = new String[cmdlst.size()];
-        cmdarr = cmdlst.toArray(cmdarr);
+        final Pair<String, String> pair = this.flagAndUser(usr);
+        cmdlst.add(pair.getLeft());
+        cmdlst.add(pair.getRight());
```
olegmoz, line 171: @genryxy when I said about in-lining these methods I meant that we can just put an `if {...}` here and it would be more simple and less verbose:
```java
if (user.isPresent()) {
  cmdlst.add("--user");
  cmdlst.add(String.format(...));
}
```
or even
```java
user.ifPresent(
  usr -> {
    cmdlst.add("--user");
    cmdlst.add(String.format(...));
  }
)
```

olegmoz (changes requested):

olegmoz (approved):
@genryxy thanks, merging

0crat:
Job `gh:artipie/artipie#639` is not assigned, can't get performer

Issue #406, opened 2020-07-30
Github authentication error should not cause 500 error


g4s8:
When using Github authentication, auth errors from Github API causes 500 error for user. We should handle authentication errors and return proper status code in that case 401.

```
artipie         | Caused by: java.lang.AssertionError: HTTP response status is not equal to 200:
artipie         | 401 Unauthorized [https://api.github.com/user]
artipie         | Server: GitHub.com
artipie         | Date: Thu, 30 Jul 2020 14:39:46 GMT
artipie         | Content-Type: application/json; charset=utf-8
artipie         | Content-Length: 83
artipie         | Status: 401 Unauthorized
artipie         | X-GitHub-Media-Type: github.v3
artipie         | X-RateLimit-Limit: 60
artipie         | X-RateLimit-Remaining: 55
artipie         | X-RateLimit-Reset: 1596123034
artipie         | Access-Control-Expose-Headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, Deprecation, Sunset
artipie         | Access-Control-Allow-Origin: *
artipie         | Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
artipie         | X-Frame-Options: deny
artipie         | X-Content-Type-Options: nosniff
artipie         | X-XSS-Protection: 1; mode=block
artipie         | Referrer-Policy: origin-when-cross-origin, strict-origin-when-cross-origin
artipie         | Content-Security-Policy: default-src 'none'
artipie         | Vary: Accept-Encoding, Accept, X-Requested-With
artipie         | 
artipie         | {"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"}
artipie         | Expected: HTTP response with status 200
artipie         |      but: was <401 Unauthorized [https://api.github.com/user]
artipie         | Server: GitHub.com
artipie         | Date: Thu, 30 Jul 2020 14:39:46 GMT
artipie         | Content-Type: application/json; charset=utf-8
artipie         | Content-Length: 83
artipie         | Status: 401 Unauthorized
artipie         | X-GitHub-Media-Type: github.v3
artipie         | X-RateLimit-Limit: 60
artipie         | X-RateLimit-Remaining: 55
artipie         | X-RateLimit-Reset: 1596123034
[...]

PR #640, opened 2020-09-27
Issue #406: Fix Catch 401 result from GitHub

IlyaMoskva:
#406 Must be fixed


olegmoz (changes requested):
@IlyaMoskva thanks for the PR. The PR lacks any tests to reproduce the issue and to ensure that the issue is fixed. Also I do not understand the design choice to handle the exception in `SliceByPath`. This class is designed for routing requests by path to repositories and has nothing to do with auth or GitHub auth

Issue #647, opened 2020-10-01
Expand readme


olenagerasimova:
We lack some information in our readmes, at least: 
- repo permissions descriptions
- info about storages types and configuration

Let's add table of content to the readme after quick start section along with missing documentation.

0crat:
Job `gh:artipie/artipie#647` is not assigned, can't get performer

PR #648, opened 2020-10-01
#647 - document repo perms

olenagerasimova:
Part of #647 
Added TOC and documented repo permissions. You can check md [here](https://github.com/olenagerasimova/artipie/tree/647-repo-perms)

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ docker run -d -v /var/artipie:/var/artipie` -p 80:80 \
 
 ## Repository permissions
 
-Permissions for repository operations can granted in the repo configuration file:
+Permissions for repository operations can be granted in the repo configuration file:
 ```yaml
 repo:
   ...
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,6 +71,14 @@ We recommend you read the "Architecture" section in our
 [White Paper](https://github.com/artipie/white-paper) to fully
 understand how Artipie is designed.
 
+## Contents
+
+- [Multitenancy](#multitenancy)
+- [Repository permissions](#repository-permissions)
+- [Storage configuration](#storage-configuration)
+- [Metrics](#metrics)
+- [Artipie REST API](#artipie-rest-api)
```
olegmoz, line 80: @olenagerasimova I think we need some adjustements to order of items according to importance of feature to users:

- "Storage configuration" should come first - it's most important part, every user will need to configure storage I believe.
- Then I'd put "Repository permissions" before "Multitenancy", because configuring permissions is usually required when there more then one team is working with Artipie or it is available to public users from outside of organization. "Multitenancy" is feature for even larger installations when there are even there are independent parties are using the same Artipie installation, such as `central.artipie.com`.

So the following order in my opinion will make more sense:

- Storage configuration
- Repository permissions
- Multitenancy
- Metrics
- Artipie REST API

olegmoz (changes requested):
@olenagerasimova thanks, please see some comments

olenagerasimova:
@olegmoz thanks, all corrected

olegmoz (approved):

Issue #602, opened 2020-09-17
NuGet integration tests


olegmoz:
Add integration tests that verify that NuGet can upload and install packages with authentication enabled and disabled.
Such integration tests are present in NuGet adapter repository

0crat:
You @olegmoz[/z](https://www.0crat.com/u/olegmoz) need to have one of these roles in order to do "Request order start": `ARC`, `PO`; I'm sorry to say this, but at the moment you've got no roles in this project

olegmoz:
It is problematic to test authentication for NuGet, see https://github.com/artipie/artipie/issues/657

0pdd:
@olegmoz the puzzle [#677](https://github.com/artipie/artipie/issues/677) is still not solved.

0pdd:
@olegmoz 2 puzzles [#677](https://github.com/artipie/artipie/issues/677), [#679](https://github.com/artipie/artipie/issues/679) are still not solved.

0crat:
Job `gh:artipie/artipie#602` is not assigned, can't get performer

0pdd:
@olegmoz all 2 puzzles are solved here: [#677](https://github.com/artipie/artipie/issues/677), [#679](https://github.com/artipie/artipie/issues/679).

PR #653, opened 2020-10-02
#602 - push package NugetITCase

genryxy:
Part of #602 
Verify that pushing package with enable authentication is working.


src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @Test
+    void shouldPushPackage() throws Exception {
+        MatcherAssert.assertThat(
+            this.pushPackage(),
+            new StringContains("Your package was pushed.")
+        );
+    }
+
+    private void nugetConfig() throws Exception {
```
olegmoz, line 102: @genryxy this method does not return anything, but does an action: creates NuGet config file. It should be renamed to indicate that

src/test/java/com/artipie/nuget/NewtonJsonResource.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+package com.artipie.nuget;
+
+import com.google.common.io.ByteStreams;
+import org.cactoos.io.ResourceOf;
+
+/**
+ * Newton.Json package resource.
+ *
+ * @since 0.1
```
olegmoz, line 32: @genryxy `@since` tag value is off, it should be `0.12` now

src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -0,0 +1,158 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.testcontainers.Testcontainers;
+import org.testcontainers.containers.GenericContainer;
+
+/**
+ * Integration tests for Nuget repository.
+ * @since 0.11
```
olegmoz, line 43: @genryxy  `@since` tag value is off, it should be 0.12 now

olegmoz (changes requested):
@genryxy thanks, please see come comments

genryxy:
@olegmoz thanks, corrected

src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -43,6 +45,7 @@
  * @since 0.12
  */
 @SuppressWarnings("PMD.AvoidDuplicateLiterals")
+@EnabledOnOs({OS.LINUX, OS.MAC})
```
olegmoz, line 48: @genryxy probably you need to enable the test on windows only, not enable it unix-like systems

olegmoz (changes requested):
@genryxy please fix the build, it seems that it just hangs on Ubuntu

src/test/java/com/artipie/nuget/NugetITCase.java
genryxy: @olegmoz in the previous commit it fails on windows https://github.com/artipie/artipie/runs/1197922355#step:5:7796

olegmoz: @genryxy that's because `testcontainers` library does not support Windows. That is sad. It means you need to use a Linux image instead of Windows one

olegmoz: @genryxy you may find such images here: https://hub.docker.com/_/microsoft-dotnet-core

genryxy: @olegmoz in this test `mcr.microsoft.com/dotnet/sdk:5.0` image is used

olegmoz (approved):
@genryxy thanks, merging

0crat:
Job `gh:artipie/artipie#653` is not assigned, can't get performer

Issue #660, opened 2020-10-06
Extract class to build repo config yaml


olenagerasimova:
A lot of tests (all the adapters ITs) classes require repo config yaml to be build. Now code to do that is repeating in each class, let's create another class in the test scope to avoid duplications.

0crat:
@olegmoz[/z](https://www.0crat.com/u/olegmoz) the issue is closed not by @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) (its creator); I won't close the order; please, re-open it and ask @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) to close it

PR #673, opened 2020-10-08
#660 - use and extend RepoConfigYaml

olenagerasimova:
Part of #660 
Used `RepoConfigYaml` in couple more tests and extended it with method to save config to provided storage. I'm planning to use `RepoConfigYaml#saveTo` method instead of `RepoPerms#saveSettings` as `RepoPerms` class should work with permission and patterns only, not with the whole repo config.


olegmoz (approved):

0crat:
Job `gh:artipie/artipie#673` is not assigned, can't get performer

PR #675, opened 2020-10-08
#602 - Verify install nugetIT

genryxy:
Part of #602 
Verify that installing pushed package with enable authentication is working.


src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -69,16 +71,22 @@
...
+    private String pckg;
 
     @BeforeEach
     void init() throws Exception {
         final String name = "my-nuget";
         this.server = new ArtipieServer(this.tmp, name, this.config());
-        this.port = this.server.start();
-        Testcontainers.exposeHostPorts(this.port);
+        final int port = this.server.start();
+        this.url = String.format("http://host.testcontainers.internal:%d/my-nuget", port);
```
olegmoz, line 88: @genryxy `this.url` field is initialized only after `this.config()` method call. But it is used inside `config()` method. Not sure how this test works.

src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -69,16 +71,22 @@
...
-     * Artipie server port.
+     * URL 'http://host:port/my-nuget'.
      */
-    private int port;
+    private String url;
+
+    /**
+     * Package name in the temporary directory.
+     */
+    private String pckg;
```
olegmoz, line 81: @genryxy value for this field is always `"newtonsoft.json"`, so this field should not be mutable. Please replace it with constant

src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -106,10 +114,28 @@ void shouldPushPackage() throws Exception {
         );
     }
 
-    private void createNugetConfig() throws Exception {
-        final String url = String.format(
-            "http://host.testcontainers.internal:%d/my-nuget", this.port
+    // @checkstyle MagicNumberCheck (2 lines)
```
olegmoz, line 117: @genryxy I think we may move this warning to class level, as both test methods in this class have `@Timeout` annotations

src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -118,8 +144,14 @@ private void createNugetConfig() throws Exception {
                 "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n",
                 "<configuration>",
                 "<packageSources>",
-                String.format("<add key=\"%s\" value=\"%s\"/>", source, url),
+                String.format(
+                    "<add key=\"%s\" value=\"%s\"/>",
+                    source, String.format("%s/index.json", this.url)
+                ),
```
olegmoz, line 150: @genryxy here we specify URL for `"artipie-nuget-test"` alias. But I am not sure we are actually using it anywhere, can't we just remove this `<packageSources>` section?

src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -139,15 +169,14 @@ private String exec(final String... command) throws Exception {
...
         Files.write(
-            this.tmp.resolve(pckg),
+            this.tmp.resolve(this.pckg),
             new NewtonJsonResource("newtonsoft.json.12.0.3.nupkg").bytes()
         );
         return this.exec(
-            "dotnet", "nuget", "push", pckg, "--api-key", "this.source",
-            "-s", String.format(uri, this.port)
+            "dotnet", "nuget", "push", this.pckg,
+            "-s", String.format("%s/index.json", this.url)
```
olegmoz, line 179: @genryxy file name you pass to `nuget push` command does not matter. NuGet repository extracts package name from it's content. So I'd suggest to keep using `UUID.randomUUID().toString()` for filename same as it was

olegmoz (changes requested):

src/test/java/com/artipie/nuget/NugetITCase.java
genryxy: @olegmoz it works normal since I'm recording file with config after the port number is known

genryxy:

olegmoz (changes requested):
@genryxy please see response about port problem

genryxy:
@olegmoz thanks, fixed problem with port as you suggested

src/test/java/com/artipie/nuget/NugetITCase.java
```diff
@@ -73,17 +79,19 @@
...
         final String name = "my-nuget";
-        this.server = new ArtipieServer(this.tmp, name, this.config());
-        final int port = this.server.start();
-        this.url = String.format("http://host.testcontainers.internal:%d/my-nuget", port);
-        Testcontainers.exposeHostPorts(port);
+        try (ServerSocket socket = new ServerSocket(0)) {
+            this.port = socket.getLocalPort();
+            this.server = new ArtipieServer(this.tmp, name, this.config());
+        }
+        this.port = this.server.start();
```
olegmoz: @genryxy here you are using found free port to put it into config. What I meant is to get the port and start `ArtipieServer` on it. This way you will have correct port in config from the start and will not need to overwrite config later

olegmoz (changes requested):

src/test/java/com/artipie/nuget/NugetITCase.java
genryxy: @olegmoz  in this case I need to add new constructor with four params in `ArtipieServer`? Or is there a better solution that I don't see?

olegmoz: @genryxy ah, I see. Sorry thought there is option to run `ArtipieServer` on specified port at the moment. Could you please create a puzzle about it? Then we will be good to merge

genryxy:
@olegmoz thanks, added todo

olegmoz (approved):
@genryxy thanks, merging

0crat:
Job `gh:artipie/artipie#675` is not assigned, can't get performer

PR #686, opened 2020-10-12
#633 - Response 404 to request not related to repos

### LINKED ISSUE #633: Respond with 404 to request not related to repositories
`SliceByPath` should not crash, but respond with `404` for requests that cannot be directed to a repository, such as `/favicon.ico` or `robots.txt` for example

genryxy:
Closes #633 
Respond with 404 to request not related to repositories in `SliceByPath`

### DIFF
--- a/src/main/java/com/artipie/http/SliceByPath.java
+++ b/src/main/java/com/artipie/http/SliceByPath.java
@@ -108,7 +108,7 @@ final class SliceByPath implements Slice {
                 key = new Key.From(split[0], split[1]);
             }
         } else {
-            if (split.length == 1) {
+            if (split.length >= 1) {
                 key = new Key.From(split[0]);
             }
         }
--- a/src/main/java/com/artipie/http/SliceByPath.java
+++ b/src/main/java/com/artipie/http/SliceByPath.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
+import java.util.Optional;
 import org.reactivestreams.Publisher;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/http/SliceByPath.java
```diff
@@ -76,18 +76,13 @@ public Response response(final String line, final Iterable<Map.Entry<String, Str
...
-                    throw new IllegalStateException("Expected at least 2 path segments");
-                }
-                key = new Key.From(split[0], split[1]);
-            } else {
-                if (split.length < 1) {
-                    throw new IllegalStateException("Expected at least 1 path segment");
-                }
-                key = new Key.From(split[0]);
+            key = this.keyFromPath(new RequestLineFrom(line).uri().getPath());
+            if (key.equals(Key.ROOT)) {
```
olegmoz, line 80: @genryxy I see that you use `Key.ROOT` as special value that means repository name cannot be extracted from path. That is not that great design to have such values with special handling. It will be more obvious to return `Optional<Key>` from method and have special handling for absent value

src/main/java/com/artipie/http/SliceByPath.java
```diff
@@ -98,4 +93,25 @@ public Response response(final String line, final Iterable<Map.Entry<String, Str
...
+        if (this.settings.layout().equals("org")) {
+            if (split.length >= 2) {
+                key = new Key.From(split[0], split[1]);
+            }
+        } else {
+            if (split.length >= 1) {
+                key = new Key.From(split[0]);
+            }
+        }
+        return key;
```
olegmoz, line 115: @genryxy using not `final` variables should be avoided. Mutable values make code harder to understand. Please change `key` variable to be `final`.

olegmoz (changes requested):

olegmoz (approved):
@genryxy thanks, merging

0crat:
Job `gh:artipie/artipie#686` is not assigned, can't get performer

Issue #690, opened 2020-10-14
Create class to compose credentials yaml config


olenagerasimova:
Let's create class to build credentials yaml config in test scope to replace multiple repeating places where this config is created.

0crat:
@olegmoz[/z](https://www.0crat.com/u/olegmoz) the issue is closed not by @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) (its creator); I won't close the order; please, re-open it and ask @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) to close it

Issue #203, opened 2020-06-11
credentials.yml may be simplied


yegor256:
Currently, this is the format:

```yaml
credentials:
  jane:
    pass: "plain:qwerty"
  john:
    pass: "sha256:xxxxxxxxxxxxxxxxxxxxxxx"
```

I'm suggesting this:

```yaml
credentials:
  jane:
    type: plain
    pass: qwerty
  john:
    type: sha256
    pass: xxxxxxxxxxxxxxxxxxxxxxx
```

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8), @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) please, pay attention to this issue

0crat:
The job [#203](https://github.com/artipie/artipie/issues/203) assigned to @paulodamaso[/z](https://www.0crat.com/u/paulodamaso), here is [why](https://www.0crat.com/footprint/CT2E6TK9B/fe5ade0c-9958-4110-83f6-1d3c58c305d4); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be a monetary reward for this job

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) this job was assigned to you 5days ago. It will be taken away from you soon, unless you close it, see [§8](http://www.zerocracy.com/policy.html#8). Read [this](http://www.yegor256.com/2014/04/13/no-obligations-principle.html) and [this](http://www.yegor256.com/2014/11/24/principles-of-bug-tracking.html), please.

paulodamaso:
@yegor256 merged, please close

g4s8:
@paulodamaso PR #215 was merged, thanks.

PR #695, opened 2020-10-15
#690 - use CredsConfigYaml

olenagerasimova:
Part of #690 
In #203 we introduced config format with separate yaml field for password type: 
```
username:
  type: plain/sha256
  pass: 123
...
``` 
Such format seems more readable, that's why I've corrected `CredsConfigYaml` to use separate field for type and left old format only in `AuthFromYamlTest` to verify it's working. Also added several methods to `CredsConfigYaml` and used it tests.


olegmoz (approved):

0crat:
Job `gh:artipie/artipie#695` is not assigned, can't get performer

PR #699, opened 2020-10-16
#690 - use CredsConfigYaml, correct Users.FromStorageYaml and readme

### LINKED ISSUE #690: Create class to compose credentials yaml config

olenagerasimova:
Closes #690 
Corrected `Users.FromStorageYaml` to save password with format in separate field and used `CredsConfigYaml` in the corresponding test. Also, corrected and extended readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,15 +167,18 @@ meta:
 
 If the `type` is set to `file`, another YAML file is required in the storage, with
 a list of users who will be allowed to create repos
-(each `pass` is combination or either `plain` or `sha256` and a text):
+(`type` is password format, `plain` and `sha256` types are supported):
 
 ```yaml
 credentials:
   jane:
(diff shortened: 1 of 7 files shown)


olegmoz (approved):
@olenagerasimova thanks, looks good

Makes me think that user configuration format needs some improvement:
```
<username>:
  type: ...
  pass: ...
```
looks like there is `type` of user. It is unclear that `type` is actually password type. Maybe `pass` attribute may be string OR mapping:
```
<username>
  pass: "plain:qwerty"
```
and another option
```
<username>
  pass: 
    type: plain
    value: qwerty
```
WDYT?

0crat:
Job `gh:artipie/artipie#699` is not assigned, can't get performer

Issue #738, opened 2020-11-05
Fix path handling for NpmProxy


genryxy:
If in `SliceFromConfig`  use `""` instead of `cfg.path()` then `meta.json` will be found by proxy and loaded but `.tgz` will not be found (output below). If use `cfg.path()` then neither of the two files will be found:
```
            case "npm-proxy":
                slice = trimIfNotStandalone(
                    settings, standalone,
                    new NpmProxySlice(
//                        cfg.path(),
                        "",
                        new NpmProxy(
                            new NpmProxyConfig(cfg.settings().orElseThrow()),
                            Vertx.vertx(),
                            cfg.storage()
                        )
                    )
                );
                break;
```

```
[INFO] main com.artipie.VertxMain: Artipie was started on port 65447
[INFO] FileStorage-3 com.artipie.asto.fs.FileStorage: Found 4 objects by the prefix "" in C:\Users\AWX971~1\AppData\Local\Temp\junit13188099467082796481\origin\repos by C:\Users\AWX971~1\AppData\Local\Temp\junit13188099467082796481\origin\repos: [_credentials.yml, npm-origin.yaml, npm-origin/@hello/simple-npm-project/-/@hello/simple-npm-project-1.0.1.tgz, npm-origin/@hello/simple-npm-project/meta.json]
[DEBUG] FileStorage-5 com.artipie.RepoConfig: parsed yaml config:
repo:
  type: npm
  storage:
    type: fs
    path: "C:\Users\AWX971~1\AppData\Local\Temp\junit13188099467082796481\origin\repos"
  url: "http://localhost:65447/npm-origin"
[INFO] main com.artipie.VertxMain: Artipie was started on port 65482
[INFO] FileStorage-6 com.artipie.asto.fs.FileStorage: Found 2 objects by the prefix "" in C:\Users\AWX971~1\AppData\Local\Temp\junit13188099467082796481\proxy\repos by C:\Users\AWX971~1\AppData\Local\Temp\junit13188099467082796481\proxy\repos: [_credentials.yml, my-npm.yaml]
[DEBUG] FileStorage-7 com.artipie.RepoConfig: parsed yaml config:
repo:
  type: "npm-proxy"
  storage:
    type: fs
    path: "C:\Users\AWX971~1\AppData\Local\Temp\junit13188099467082796481\proxy\repos"
  url: "http://host.testcontainers.internal:65482/my-npm/"
  path: "my-npm"
  settings:
    remote:
[...]

PR #739, opened 2020-11-05
#712 - Add npmProxyIT

### LINKED ISSUE #712: NPM proxy integration tests
Add integration tests that verify that NPM proxy can install packages

genryxy:
Closes #712 
Now test is disabled. See #738


src/test/java/com/artipie/npm/NpmProxyITCase.java
```diff
@@ -0,0 +1,217 @@
...
+                "http://localhost:%d/%s/@hello/simple-npm-project/-/@hello/simple-npm-project-1.0.1.tgz", this.origin.port(), NpmProxyITCase.ORIGIN
+            )
+        ).openConnection();
+        con.setRequestMethod("GET");
+        con.setDoOutput(true);
+        MatcherAssert.assertThat(
+            "Verify that tgz file exists. Response status is 200",
+            con.getResponseCode(),
+            new IsEqual<>(Integer.parseInt(RsStatus.OK.code()))
+        );
```
olegmoz, line 115: @genryxy not sure why do we need this check. The test checks that `npm install` works later, this check does not look to be required

olegmoz (changes requested):

src/test/java/com/artipie/npm/NpmProxyITCase.java
genryxy: @olegmoz thanks, I wanted to show by this check that the file can actually be obtained from origin repo. Probably I should remove this check or write about removing in the created issue related to enabling test.

olegmoz: @genryxy if it helps to diagnose the problem, then maybe it deserves to be separate test, which is enabled.

genryxy: @olegmoz it rather checks that the test is written correctly, but does not help to diagnose the problem

olegmoz: @genryxy still looks like it should be moved to another test. So we will have test that checks Artipie server may be used as target for `npm install` command and separate test to check that HTTP endpoint works properly. These are different features basically

genryxy:
@olegmoz  thanks, divided the test into two

olegmoz (approved):
@genryxy thanks, merging

0crat:
Job `gh:artipie/artipie#739` is not assigned, can't get performer

Issue #606, opened 2020-09-17
Gem integration tests


olegmoz:
Add integration tests that verify that Gem can upload and install packages with authentication enabled and disabled

PR #758, opened 2020-11-17
#606 - Verify push and install for Gem

genryxy:
Part of #606 
Verify push and install for Gem without authentication


src/test/java/com/artipie/gem/package-info.java
```diff
@@ -0,0 +1,30 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/**
+ * Tests for gem-adapter related classes.
```
olegmoz, line 26: @genryxy I think it's better to say that we are testing Gem repository rather then `gem-adapter` related classes

src/test/java/com/artipie/gem/GemITCase.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @AfterEach
+    void tearDown() {
+        this.server.stop();
+        this.cntn.close();
+    }
+
+    private void push(final boolean anonymous) throws Exception {
+        final String key = new Base64Encoded("usr:pwd").asString();
```
olegmoz, line 124: @genryxy could you please explain how `key` is used? We build it from arbitrary data, but I do not see how this data is translated into user on server. Does the test fails if auth is enabled on server, but the key is not provided on push?

src/test/java/com/artipie/gem/GemITCase.java
```diff
@@ -0,0 +1,160 @@
...
+        new TestResource(String.format("gem/%s", GemITCase.RAILS))
+            .saveTo(this.storage, new Key.From(GemITCase.RAILS));
+        final String urlwithslash = this.url.string(anonymous);
```
olegmoz, line 127: @genryxy the name does not make much sense, why URL is with slash? Maybe `urlwithuser` or just `url` would be better?

olegmoz (changes requested):

src/test/java/com/artipie/gem/GemITCase.java
genryxy: @olegmoz in this case I want to show that `this.url.string(anonymous)` returns url with slash and request looks like `POST /my-gem//api/v1/gems HTTP_1_1`. So, I take substring of `urlwithslash`

src/test/java/com/artipie/gem/GemITCase.java
genryxy: @olegmoz yes, it fails because the `gem push` expects that the gem server always requires credentials. Please, see https://stackoverflow.com/questions/61384113/how-to-push-a-gem-without-auth I'll replace `usr:pwd` with `any:any` to make it little clearer.

genryxy:
@olegmoz thanks, corrected comment in `package-info`, replaced `usr:pwd` with `any:any` and answered

src/test/java/com/artipie/gem/GemITCase.java
olegmoz: @genryxy but `this.url` contains these slashes. How calling `string` method on `this.url` transforms it into URL with slashes (adds slashes)?

genryxy: @olegmoz `string` method does not add slash at the end. It just concatenates as `String.format("http://%s", this.hostportrepo)`. `hostportrepo` is initialized in the constructor
```
public RepositoryUrl(final int port, final String reponame) {
    this.hostportrepo = String.format("host.testcontainers.internal:%d/%s/", port, reponame);
}
```

olegmoz: @genryxy exactly my point, it does not add slashes. Then why code reads "apply method `string` to `this.url` with parameter `anonymous` and get URL with slashes as result`, does it make sense?

genryxy: @olegmoz thanks, renamed variable

genryxy:
@olegmoz after merging `upstream/master` in my branch `GemITCase` started failing with `com.artipie.asto.ValueNotFoundException: No value for key: my-gem.yaml.yaml`

olegmoz (approved):
@genryxy thanks, all is good now. I see that all checks are passing here, when do you see this error with my-gem.yaml?

genryxy:
@olegmoz after merging I have this problem for all ITCases. Probably, it is due  to the local problems

genryxy:
@olegmoz it seems that ITCases fail https://github.com/artipie/artipie/pull/758/checks?check_run_id=1423006614#step:5:1338

olegmoz:
@genryxy wow, thanks for pointing this out. I will try to fix it ASAP

Issue #680, opened 2020-10-09
Extract profile for IT cases


olenagerasimova:
Let's extract integration tests into separate profile and add separate action for it.

0crat:
@olegmoz[/z](https://www.0crat.com/u/olegmoz) the issue is closed not by @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) (its creator); I won't close the order; please, re-open it and ask @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) to close it

0crat:
@olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) resigned from [#680](https://github.com/artipie/artipie/issues/680), since the job is not in scope anymore

PR #764, opened 2020-11-19
#680 - Fix for integration tests action

olegmoz:
#680
Failing integration tests should cause GitHub check to fail


olegmoz:
@olenagerasimova FYI, integration tests run results were ignored by Maven because target `verify` goal wasn't specified for failsafe plugin. Still not sure why test were running in this case

PR #767, opened 2020-11-20
#606 - Add authentication push and install IT for Gem

### LINKED ISSUE #606: Gem integration tests

genryxy:
Closes #606 
Add authentication push and install IT for Gem. Now these test cases are disabled. Add todo to enable them after passing of authentication and permissions would be implemented.

### DIFF
--- a/src/test/java/com/artipie/gem/GemITCase.java
+++ b/src/test/java/com/artipie/gem/GemITCase.java
@@ -39,14 +39,20 @@ import org.hamcrest.MatcherAssert;
 import org.hamcrest.core.IsEqual;
 import org.hamcrest.core.StringContains;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledOnOs;
 import org.junit.jupiter.api.condition.OS;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 /**
  * Integration tests for Gem repository.
  * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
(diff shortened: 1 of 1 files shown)


src/test/java/com/artipie/gem/GemITCase.java
```diff
@@ -121,18 +131,25 @@ void tearDown() {
...
+                    key, GemITCase.RAILS,
+                    tmpurl.substring(0, tmpurl.length() - 1)
+                )
+            );
+        } else {
+            this.cntn.execStdout(
+                "gem", "push", GemITCase.RAILS,
+                "--host", tmpurl.substring(0, tmpurl.length() - 1)
+            );
+        }
```
olegmoz, line 152: @genryxy it seems odd that there is no `GEM_HOST_API_KEY` specified in case of non-anonymous operation

olegmoz (changes requested):

src/test/java/com/artipie/gem/GemITCase.java
genryxy: @olegmoz  I think that credentials in this case should be obtained from the url that looks like `http://user:pass@host:port/my-gem`

olegmoz: @genryxy could you please prove it by some link to tutorial or docs? Or maybe there is existing test in gem-adapter repository? I am under impression that `GEM_HOST_API_KEY` is the only authentication option supported by `gem` CLI tool

genryxy:
@olegmoz thanks, you are right. Credentials for pushing could be obtained only from `/root/.gem/credentials` or from `GEM_HOST_API_KEY ` that was set before (https://guides.rubygems.org/command-reference/#description-18)

olegmoz (approved):
@genryxy thanks, merging

PR #780, opened 2020-11-30
Simplify YamlSettings.java

portlek:


olegmoz (changes requested):
@portlek thanks for the PR. I'd suggest to remove `YamlSettings.KEY_META` field altogether, as it is now used in `meta` method only.

olegmoz (approved):
@portlek thanks, merging

Issue #768, opened 2020-11-20
Support authentication in Gem repositories


olegmoz:
Gem adapter supports authentication, but it is not supported in Artipie. It is required to add authentication support for Gem adapters and enable existing integration tests

olegmoz:
@rultor release, tag=`0.13`

PR #782, opened 2020-12-01
#768 - Support Basic auth for Gem

genryxy:
Part of #768 
Updated version of gem-adapter and http module. Enabled test for this feature.


olegmoz (approved):
@genryxy thanks, merging

PR #785, opened 2020-12-02
#586 - Layout extracted

### LINKED ISSUE #586: Extract class for layout
Now `layout` is represented with `String` in `Settings`. In many places the values is read and handled manually according to value `org` or `flat`. These two types of layout could be extracted to classes with same interface that parses repo name from URI and does other required operations

olegmoz:
Closes #586
Layout extracted


genryxy (approved):
@olegmoz thanks, approve

Issue #778, opened 2020-11-27
Increase test coverage for classes that parse yaml files


genryxy:
It would be good to increase test coverage for classes that parse yaml files

- [x] `RepoConfig` (#779 )
- [x]  `StorageAliases.FromYaml` and `RepositioriesFromStorage` (#783 #792)
- [x] `YamlProxyConfig` (#794)

0crat:
Job `gh:artipie/artipie#778` is not assigned, can't get performer

PR #794, opened 2020-12-04
#778 - Add tests for YamlProxyConfig, extract class StorageYamlConfig

genryxy:
Part of #778 
Added negative tests for `YamlProxyConfig`. Also extracted class `StorageYamlConfig` because `RepoConfig` and  `YamlProxyConfig` use the same logic.


src/test/java/com/artipie/repo/StorageYamlConfigTest.java
```diff
@@ -0,0 +1,111 @@
...
+            new StorageYamlConfig(
+                Yaml.createYamlScalarBuilder().addLine(alias).buildPlainScalar(),
+                new StorageAliases.FromYaml(
+                    Yaml.createYamlMappingBuilder().add(
+                        "storages", Yaml.createYamlMappingBuilder()
+                            .add(
+                                alias, Yaml.createYamlMappingBuilder()
+                                    .add(StorageYamlConfigTest.TYPE, "fs")
+                                    .add(StorageYamlConfigTest.PATH, "/some/path")
+                                    .build()
```
olegmoz, line 84: @genryxy it seems that this code for creating YAML with `type` and `path` is duplicated. I'd suggest to extract method for it and remove static fields `TYPE` and `PATH` as they will be used only once

src/test/java/com/artipie/repo/StorageYamlConfigTest.java
```diff
@@ -0,0 +1,111 @@
...
+        );
+    }
+
+    @Test
+    void returnsSubStorage() {
+        MatcherAssert.assertThat(
+            this.config().subStorage(new Key.From()),
+            new IsInstanceOf(Storage.class)
+        );
+    }
```
olegmoz, line 99: @genryxy I think it should be tested that in fact `SubStorage` is returned. It would be even better to ensure that `prefix` passed to method is applied correctly

olegmoz (changes requested):

genryxy:
@olegmoz thanks, added test to verify prefix of `SubStorage` and extracted variable

olegmoz (approved):
@genryxy thanks, merging

PR #795, opened 2020-12-04
Change path of default resources.

portlek:
This pull request will fix this error which shows up when the configuration files not found and if Artipie itself creates the default files from its classpath.
image

### DIFF
--- a/src/main/java/com/artipie/VertxMain.java
+++ b/src/main/java/com/artipie/VertxMain.java
@@ -195,6 +195,7 @@ public final class VertxMain {
         final Collection<RepoConfig> configs = storage.list(Key.ROOT).thenApply(
             keys -> keys.stream().map(Key::string)
                 .filter(name -> name.endsWith(yaml))
+                .filter(name -> !name.startsWith("_"))
                 .map(name -> name.substring(0, name.length() - yaml.length()))
                 .map(name -> new RepositoriesFromStorage(storage).config(name))
                 .map(stage -> stage.toCompletableFuture().join())
--- a/src/main/resources/example/repo/_credentials.yaml
+++ b/src/main/resources/example/repo/_credentials.yaml
@@ -1,3 +1,3 @@
-credentials:
+repo:
   artipie:
     pass: plain:artipie
--- a/src/main/resources/example/repo/_credentials.yaml
+++ b/src/main/resources/example/repo/_credentials.yaml
@@ -1,3 +1,3 @@
-repo:
+credentials:
   artipie:
     pass: plain:artipie
--- a/src/main/resources/example/repo/_permissions.yaml
+++ b/src/main/resources/example/repo/_permissions.yaml
@@ -1,3 +1,3 @@
-permissions:
+repo:
   artipie:
     - api
--- a/src/main/resources/example/repo/_permissions.yaml
+++ b/src/main/resources/example/repo/_permissions.yaml
@@ -1,3 +1,3 @@
-repo:
+permissions:
   artipie:
     - api
--- a/src/main/resources/example/repo/_storages.yaml
+++ b/src/main/resources/example/repo/_storages.yaml
@@ -1,4 +1,4 @@
-storages:
+repo:
   default:
     type: fs
     path: ./.storage/data
--- a/src/main/resources/example/repo/_storages.yaml
+++ b/src/main/resources/example/repo/_storages.yaml
@@ -1,4 +1,4 @@
-repo:
+storages:
   default:
     type: fs
     path: ./.storage/data


genryxy:
@portlek thanks for your PR, but the changes are incorrect. Please, see this article https://blog.artipie.com/2020-11-06-configure-artipie-maven-repo.html to understand what config files are obligatory. This exception looks like the `repo-name.yaml` file is absent. In this article the file is called `my-maven.yaml`

portlek:
> @portlek thanks for your PR, but the changes are incorrect. Please, see this article https://blog.artipie.com/2020-11-06-configure-artipie-maven-repo.html to understand what config files are obligatory. This exception looks like the `repo-name.yaml` file is absent. In this article the file is called `my-maven.yaml`

Hmm, so it means there is a bug about that because the credentials, permissions, storages files must start with `repo:` but it shouldn't be.
If I don't change my starting of the files as `repo:` it shows the error. So I assumed that there is a bug.

genryxy:
@portlek There is no bug because we have integration tests with authentication for adapters in `artipie` module. See https://github.com/artipie/artipie/blob/master/src/test/java/com/artipie/npm/NpmITCase.java or https://github.com/artipie/artipie/blob/master/src/test/java/com/artipie/pypi/PypiITCase.java

portlek:
> @portlek There is no bug because we have integration tests with authentication for adapters in `artipie` module. See https://github.com/artipie/artipie/blob/master/src/test/java/com/artipie/npm/NpmITCase.java or https://github.com/artipie/artipie/blob/master/src/test/java/com/artipie/pypi/PypiITCase.java

So why I'm getting that error? I just run the default .jar file and It created the default resource files then I'm getting that error. I examined the artipie's code and I saw(if I'm not wrong) **each** files must start with `repo:`

portlek:
image
It tries getting `repo:` key for each file.

portlek:
```java
        .filter(name -> name.endsWith(yaml))
        .map(name -> name.substring(0, name.length() - yaml.length()))
        .map(name -> new RepositoriesFromStorage(storage).config(name))
        .map(stage -> stage.toCompletableFuture().join())
        .collect(Collectors.toList())
).toCompletableFuture().join();
```
This code must have one more filter that filters if the file name does not start with `_`

portlek:
Anyway, I will find the problem here.

genryxy:
@portlek probably you have another directory structure? For example `_credentials.yaml` instead of `_credentials.yml`
```
.
├── artipie.yaml
├── configs
    ├── my-maven.yaml
    └── _credentials.yml
```

portlek:
> @portlek probably you have another directory structure? For example `_credentials.yaml` instead of `_credentials.yml`
> 
> ```
> .
> ├── artipie.yaml
> ├── configs
>     ├── my-maven.yaml
>     └── _credentials.yml
> ```
> 
> `artipie.yaml` in this case contains
> 
> ```
> meta:
>   storage:
>     type: fs
>     path: /var/artipie/configs
>   credentials:
>     type: file
>     path: _credentials.yml
> ```

The default resources are not like that, they end with `.yaml` instead of `.yml` also artipie.yaml contains `_credentials.yaml` instead of `_credentials.yml`

portlek:
[layout.zip](https://github.com/artipie/artipie/files/5652339/layout.zip)
my layout is like that also I have an example maven project to upload which has this pom.xml:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="http://maven.apache.org/POM/4.0.0"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.example</groupId>
    <artifactId>artipieTest</artifactId>
    <version>1.0</version>

    <properties>
        <maven.compiler.source>11</maven.compiler.source>
        <maven.compiler.target>11</maven.compiler.target>
    </properties>
   
    <distributionManagement>
        <repository>
            <id>artipie</id>
            <url>http://localhost:8080/my-maven</url>
        </repository>
    </distributionManagement>
</project>
```
and I always get this error:
image

I tried everything, debugs, some little changes but no

portlek:
with `flat` layout, I can upload but can't get. with `org`, I can't upload + download. And I'm using default resource, so because of that I'm saying there must be a bug.

genryxy:
@portlek thanks, for the bug found, we will fix it soon. Еhis is most likely due to configs files' extensions. Try to use `artipie.yaml` and `_credentials.yml`. For such extensions, it should now work.

PR #799, opened 2020-12-10
#798 - update management-api

### LINKED ISSUE #798: Update management-api and remove dashboard slices from this repo
Dashboard was moved to `management-api`, let's update version and remove dashboard slices and pages.

Also there are two bugs to fix:
- on start we should not try to read files with `_` prefix and repo configs (`com/artipie/VertxMain.java:197`), these files are credentials, permissions and default storage configuration
- `ArtipieApi` and `DashboardSlice` should read `permissions` sections from `_permissions.yaml` and then pass it to `YamlPermissions`

olenagerasimova:
Closes #798 
I've updated `management-api`, removed dashboard slices and pages, fixed couple of bugs.


src/main/java/com/artipie/VertxMain.java
```diff
@@ -194,7 +194,7 @@ private void startRepos(final Settings settings, final Metrics metrics) {
         final String yaml = ".yaml";
         final Collection<RepoConfig> configs = storage.list(Key.ROOT).thenApply(
             keys -> keys.stream().map(Key::string)
-                .filter(name -> name.endsWith(yaml))
+                .filter(name -> name.endsWith(yaml) && name.charAt(0) != '_')
```
olegmoz, line 197: @olenagerasimova it seems that this change is not really related to other changes in this PR

olegmoz (changes requested):
@olenagerasimova thanks, please see one comment

src/main/java/com/artipie/VertxMain.java
olenagerasimova: @olegmoz hm, please, check #798 description - I've described this bug there, so I think it can be make here

olegmoz: @olenagerasimova I see it in the issue now, yet do not understand how these bug fixes are related to replacing existing classes with api-management library. Especially the one with underscore names. If we don't allow names starting with underscore, then with should document it

olenagerasimova: @olegmoz dashboard with `org` layout cannot be used while any of these bugs are valid - I faced them trying to start artipie and access any dashboard page locally. 
I'd suggest to properly deal with underscore names (or catching exceptions with invalid repo configs, whatever we decide) in the separate issue/PR.

olegmoz: @olenagerasimova got it, thanks

olegmoz (approved):

PR #801, opened 2020-12-11
#775 - test for ArtipieApi

### LINKED ISSUE #775: Initial IT for Artipie API
Let's create minimal test for Artipie API: configure and start Artipie, then send http GET requests to `ArtipieApi` and dashboard and make sure `200 OK` is returned.

olenagerasimova:
Closes #775 
Added test from `ArtipieApi` and dashboard `GET` endpoints.

### DIFF
--- a/src/test/java/com/artipie/ArtipieServer.java
+++ b/src/test/java/com/artipie/ArtipieServer.java
@@ -156,6 +156,17 @@ public class ArtipieServer {
      *  for usage in all these places.
      */
     public int start() throws IOException {
+        return this.start("flat");
+    }
+
+    /**
+     * Starts the server.
+     *
+     * @param layout Layout
+     * @return Port the servers listening on.
+     * @throws IOException In case of error creating configs or running the server.
+     */
+    public int start(final String layout) throws IOException {
         final Path repos = this.root.resolve("repos");
         repos.toFile().mkdir();
         Files.write(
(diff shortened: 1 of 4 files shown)


src/test/java/com/artipie/ArtipieServer.java
```diff
@@ -156,6 +156,17 @@ public ArtipieServer(final Path root, final String name, final String config,
...
+    }
+
+    /**
+     * Starts the server.
+     *
+     * @param layout Layout
+     * @return Port the servers listening on.
+     * @throws IOException In case of error creating configs or running the server.
+     */
+    public int start(final String layout) throws IOException {
```
olegmoz, line 169: @olenagerasimova to me it does not seem logical that a configuration parameter is passed to `start` method. I'd suggest to specify layyou (or whole `artipie.yaml` content) when `ArtipieServer` is created. It might be too much to add in this PR, so I'd suggest to add a puzzle for that if you agree about this design issue.

src/test/java/com/artipie/api/ArtipieApiITCase.java
```diff
@@ -0,0 +1,138 @@
...
+                            "%s:%s", ArtipieServer.BOB.name(), ArtipieServer.BOB.password()
+                        ).getBytes()
+                    )
+                )
+            )
+        );
+        MatcherAssert.assertThat(
+            "Response status is 200",
+            con.getResponseCode(),
+            new IsEqual<>(Integer.parseInt(RsStatus.OK.code()))
```
olegmoz, line 115: @olenagerasimova as you are using java standard library in this method, I'd suggest to use `HttpURLConnection.HTTP_OK` instead of parsing `RsStatus.OK` value

olegmoz (changes requested):

olenagerasimova:
@olegmoz thanks, all corrected

olegmoz (approved):

PR #802, opened 2020-12-11
artipie/management-api#15 - prepare to transfer GetStorageSlice

olenagerasimova:
Part of artipie/management-api#15
I've extended `Repositories` interface with the method to get storage by repo name. The interface will be moved to `management-api` along with the slice.

### DIFF
--- a/src/main/java/com/artipie/Repositories.java
+++ b/src/main/java/com/artipie/Repositories.java
@@ -23,6 +23,7 @@
  */
 package com.artipie;
 
+import com.artipie.asto.Storage;
 import java.util.concurrent.CompletionStage;
 
 /**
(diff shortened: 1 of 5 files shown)


olegmoz (commented):
@olenagerasimova the PR looks good, though I am not sure why it is necessary to change `GetStorageSlice` in this repository. It could be added to management-api with dependency to interface instead of `RepositoriesFromStorage` class and then used here

olenagerasimova:
@olegmoz oh, sorry, this is not a good way to deal with this slice as `com.artipie.Repositories#config()` returns `CompletionStage` of `RepoConfig` and `RepoConfig` is an artipie class. I'm not sure we should extract the interface from it and move it somewhere as these are configurations, and they belong here, not in `management-api`.

Issue #797, opened 2020-12-08
Support .yaml and .yml extensions for config files


genryxy:
Now there are can be some problems with parsing yaml files because in various classes there is logic based on the configuration files' extensions. So, it is necessary to find all these places and make support for files with two extensions (`yaml` and `yml`).

0pdd:
@genryxy the puzzle [#818](https://github.com/artipie/artipie/issues/818) is still not solved.

0pdd:
@genryxy the only puzzle [#818](https://github.com/artipie/artipie/issues/818) is solved here.

PR #807, opened 2020-12-14
#797 - Check existing, trim filename for different extensions

genryxy:
Part of #797 
Check existing in storage by name for two extensions (`yaml` and `yml`). Trim filename that contains extension.


src/main/java/com/artipie/VertxMain.java
```diff
@@ -194,9 +195,10 @@ private void startRepos(final Settings settings, final Metrics metrics) {
         final String yaml = ".yaml";
         final Collection<RepoConfig> configs = storage.list(Key.ROOT).thenApply(
             keys -> keys.stream().map(Key::string)
-                .filter(name -> name.endsWith(yaml) && name.charAt(0) != '_')
-                .map(name -> name.substring(0, name.length() - yaml.length()))
-                .map(name -> new RepositoriesFromStorage(storage).config(name))
+                .filter(name -> name.charAt(0) != '_')
+                .map(name -> new ConfigFilesExtensions.Trim(name).value())
```
olegmoz, line 199: @genryxy `ConfigFilesExtensions.Trim` this class name does not look like name of an object, but looks more like a verb. I'd suggest to try come up with some better design

src/main/java/com/artipie/http/ArtipieRepositories.java
```diff
@@ -67,7 +68,7 @@ public ArtipieRepositories(final Settings settings) {
     public Slice slice(final Key name, final boolean standalone) {
         final Storage storage = this.settings.storage();
         return new AsyncSlice(
-            storage.exists(new Key.From(String.format("%s.yaml", name.string()))).thenCompose(
+            new ConfigFilesExtensions(name.toString()).exists(storage).thenCompose(
```
olegmoz, line 71: @genryxy again here "ConfigFilesExtensions" checking that "exists" does not make sense to me, I'd suggest to make class redesign

olegmoz, line 71: @genryxy please do not use `Key.toString` method here, `Key.string` should be used in business logic and keep `toString` for debug print/logging only

olegmoz (changes requested):
@genryxy thanks, please see comments. I have major concerns about class design, so didn't review all classes in details

genryxy:
@olegmoz thanks, replaced `ConfigFilesExtensions.Trim` with regex pattern. I also renamed class to `ConfigFile`.

src/main/java/com/artipie/repo/ConfigFile.java
```diff
@@ -0,0 +1,115 @@
...
+            );
+    }
+
+    /**
+     * Obtains contents from the specified storage.
+     * @return Content of the file.
+     */
+    public CompletableFuture<Content> valueFrom() {
+        throw new UnsupportedOperationException();
+    }
```
olegmoz, line 97: @genryxy I think this method does not belong to this pull request. It does not have proper implementation yet, so I'd suggest to add it later with proper usage and tests

src/main/java/com/artipie/repo/ConfigFile.java
```diff
@@ -0,0 +1,115 @@
...
+import com.artipie.asto.Key;
+import com.artipie.asto.Storage;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Supporting several config files extensions (e.g. `.yaml` and `.yml`).
```
olegmoz, line 36: @genryxy please provide better description here. Now the docs contain only details about extensions, which I think belong to `trimExtensions` method. For example `existsIn` method does nto care about extensions at all. And class instance might be created for any file name with extension or without it.

src/main/java/com/artipie/repo/ConfigFile.java
```diff
@@ -0,0 +1,115 @@
...
+    public CompletableFuture<Content> valueFrom() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Trim name of the config file.
+     * @return Filename without extensions if a filename ends with either of
+     *  the two extensions, otherwise empty.
+     */
+    public Optional<String> trimExtension() {
```
olegmoz, line 104: @genryxy right now it is not really clear what this method returns. Does it trims name and return the extension or does it trim extension or return the name? Method name is `trimExtension` yet comment states "Trim name". This is misleading.

I'd suggest to split the method into two: `isYaml` for filtering files and `name` that returns `String` without extension (or full name if it has no extension). I think it would be easier to understand this way.

olegmoz (changes requested):
@genryxy thanks for doing redesign, please see some comments

genryxy:
@olegmoz thanks, corrected

olegmoz (approved):
@genryxy thanks, merging

PR #808, opened 2020-12-15
artipie/management-api#15 - update management API and use GetStorageSlice from there

olenagerasimova:
I've updated `management-api` and used `GetStorageSlice` from there.


olegmoz (approved):

PR #814, opened 2020-12-17
#797 - Use ConfigFile where RxStorage is used for config files

genryxy:
Part of #797
Use `ConfigFile` where `RxStorageWrapper` is used

### DIFF
--- a/src/main/java/com/artipie/RepoPermissionsFromStorage.java
+++ b/src/main/java/com/artipie/RepoPermissionsFromStorage.java
@@ -31,7 +31,6 @@ import com.amihaiemil.eoyaml.YamlSequenceBuilder;
 import com.artipie.asto.Content;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
-import com.artipie.asto.rx.RxStorageWrapper;
 import com.artipie.management.RepoPermissions;
 import com.artipie.management.api.ContentAsYaml;
 import com.artipie.repo.ConfigFile;
@@ -140,7 +139,7 @@ public final class RepoPermissionsFromStorage implements RepoPermissions {
 
     @Override
     public CompletionStage<Collection<PermissionItem>> permissions(final String repo) {
-        return this.repo(RepoPermissionsFromStorage.repoSettingsKey(repo)).thenApply(
+        return this.repo(new Key.From(repo)).thenApply(
             yaml -> Optional.ofNullable(yaml.yamlMapping(RepoPermissionsFromStorage.PERMS))
         ).thenApply(
             yaml -> yaml.map(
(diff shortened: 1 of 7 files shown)


src/main/java/com/artipie/RepoPermissionsFromStorage.java
```diff
@@ -177,11 +176,11 @@ public RepoPermissionsFromStorage(final Storage storage) {
...
-        return new RxStorageWrapper(this.storage)
-            .value(key)
-            .to(new ContentAsYaml())
-            .map(yaml -> yaml.yamlMapping(RepoPermissionsFromStorage.REPO_SECTION))
-            .to(SingleInterop.get());
+        return SingleInterop.fromFuture(
+            new ConfigFile(key).valueFrom(this.storage)
+        ).to(new ContentAsYaml())
+        .map(yaml -> yaml.yamlMapping(RepoPermissionsFromStorage.REPO_SECTION))
+        .to(SingleInterop.get());
```
olegmoz, line 183: @genryxy converting `new ConfigFile(...).valueFrom(...)` into `Single` and later converting this `Single` into `CompletionStage` one again seems pointless. I'd suggest to not do it and transform `CompletionStage` into result using `thenApply`, `thenCompose` etc

src/main/java/com/artipie/UsersFromStorageYaml.java
```diff
@@ -163,14 +163,14 @@ public UsersFromStorageYaml(final Storage storage, final Key key) {
...
-            .map(YamlInput::readYamlMapping)
-            .to(SingleInterop.get());
+        return SingleInterop.fromFuture(
+            new ConfigFile(this.key).valueFrom(this.storage)
+        ).flatMap(content -> new Concatenation(content).single())
+        .map(Remaining::new)
+        .map(Remaining::bytes)
+        .map(bytes -> Yaml.createYamlInput(new String(bytes, StandardCharsets.US_ASCII)))
+        .map(YamlInput::readYamlMapping)
+        .to(SingleInterop.get());
```
olegmoz, line 173: @genryxy same concern here, I'd suggest to transform `CompletionStage` into result directly without using `Single`

olegmoz (changes requested):

genryxy:
@olegmoz thanks, but I have to use converting into `Single` because `readYamlMapping` throws an `IOException` and using converting allows to catch exception. Or it would be better to catch this exception in another way? I also found code duplication in `StorageAliases#find` and replace it using `ContentAsYaml`.

olegmoz (approved):
@genryxy ok, understand the problem with exception. I'd rather removed `IOException` from `readYamlMapping` signature. But that is not in scope of this PR

Issue #804, opened 2020-12-13
Properly handle invalid repo configs on start


olenagerasimova:
On start we read all yaml files in the settings storage and try to create repositories from them. But the settings storage can contain other yaml (for example, default storage configuration of permissions to access dashboard). Now these yamls are named starting with underscore and are files starting with underscore are ignored. This approach does seem correct and convenient, we should figure out the proper way to deal with it and document it.

portlek:
what about using JSON for settings files, and YAML for repository files? Is that a good way to deal with it?

PR #815, opened 2020-12-18
#804 - Add setting for repo configs location

olenagerasimova:
Part of #804 
Added setting for repositories configs location

### DIFF
--- a/src/main/java/com/artipie/Settings.java
+++ b/src/main/java/com/artipie/Settings.java
@@ -25,12 +25,10 @@ package com.artipie;
 
 import com.amihaiemil.eoyaml.Yaml;
 import com.amihaiemil.eoyaml.YamlMapping;
-import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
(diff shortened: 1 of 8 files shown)


src/main/java/com/artipie/Settings.java
```diff
@@ -65,6 +67,13 @@
      */
     YamlMapping meta();
 
+    /**
+     * Repo configs storage key, or, in file system storage terms, subdirectory where repo
+     * configs are located relatively to the storage.
+     * @return Repo configs key
+     */
+    Optional<Key> repoConfigs();
```
olegmoz, line 75: @olenagerasimova it seems that in every case `Key` returned by this method is composed with `storage()` method result to produce a `SubStorage`. That makes me think it would be better to return `Storage` here and put `SubStorage` creation inside

olegmoz (changes requested):
@olenagerasimova thanks, please see one comment

olenagerasimova:
@olegmoz thanks, corrected

olegmoz (approved):

PR #816, opened 2020-12-18
#797 - Support both extensions for ArtipieApi and DashboardSlice

### LINKED ISSUE #797: Support .yaml and .yml extensions for config files

genryxy:
Closes #797 
Support both extensions `.yaml` and `.yml` for `ArtipieApi` and `DashboardSlice`.


olegmoz (approved):
@genryxy thanks, merging

PR #817, opened 2020-12-18
#804 - update readme

### LINKED ISSUE #804: Properly handle invalid repo configs on start

olenagerasimova:
Closes #804 
Added info about new setting to readme


olegmoz (approved):

PR #819, opened 2020-12-21
artipie/management-api#27 - Prepare to use ConfigFile in management-api

genryxy:
Part of artipie/management-api#27
Created `ConfigFileApi` which will be passed to the instances of classes from `management-api`. This class will implement interface from `management-api`.

### DIFF
--- a/src/main/java/com/artipie/repo/ConfigFileApi.java
+++ b/src/main/java/com/artipie/repo/ConfigFileApi.java
@@ -1,0 +1,97 @@
+/*
+ * The MIT License (MIT)
+ *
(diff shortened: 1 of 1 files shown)


olegmoz (changes requested):
@genryxy I do not understand idea of this PR. New class is just a wrapper for `ConfigFile`, not an interface. How it helps with using `ConfigFile` in `management-api`? I do not get it

genryxy:
@olegmoz yes, it was probably better to create this PR after creating the interface in another module. Sorry for misleading PR.

PR #827, opened 2020-12-25
#826 - Verbose output for IT cases

### LINKED ISSUE #826: Make output for IT cases more verbose
Now  only command for execution is logged in integration tests. It would be better to make output of these tests more verbose so that the problem can identified in case of failure.

genryxy:
Closes #826 
Made output for integration tests more verbose by refactoring `TestContainer`

### DIFF
--- a/src/test/java/com/artipie/test/TestContainer.java
+++ b/src/test/java/com/artipie/test/TestContainer.java
@@ -78,9 +78,7 @@ public final class TestContainer implements AutoCloseable {
      * @throws Exception In case of exception during execution command in container.
      */
     public String execStdErr(final String... command) throws Exception {
-        final Container.ExecResult exec = this.exec(command);
-        this.logStd(exec, true);
-        return exec.getStderr().replace("\n", "");
+        return replaceBreakLine(this.exec(command).getStderr());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/test/java/com/artipie/test/TestContainer.java
```diff
@@ -106,4 +133,20 @@ public void close() {
...
+    private void logStd(final Container.ExecResult exec, final boolean error) {
+        final String outputerr = String.format("\nSTDERR:\n%s", exec.getStderr());
+        if (error) {
+            Logger.info(this, String.format("\nSTDOUT:\n%s", exec.getStdout()));
+            Logger.error(this, outputerr);
+        } else {
+            Logger.warn(this, outputerr);
+        }
+    }
+
```
olegmoz, line 151: @genryxy that looks like too complicated logic for output. I'd suggest to just print stdout and stderr after every command executed as `DEBUG` logging message and throw stderr content as exception if error code is not zero.

src/test/java/com/artipie/test/TestContainer.java
```diff
@@ -88,7 +90,32 @@ public String execStdErr(final String... command) throws Exception {
...
+        if (code != 0) {
+            this.logStd(exec, true);
+            throw new IllegalStateException(
+                String.format(
+                    "'%s' failed with %s code", String.join(" ", command), code
+                )
+            );
+        }
+        this.logStd(exec, false);
+        return exec.getStdout().replace("\n", "");
```
olegmoz, line 104: @genryxy the code for replacing `\n` with empty string is duplicated 3 times in this class, I'd suggest to extract a function for that

olegmoz (changes requested):

genryxy:
@olegmoz thanks, corrected

olegmoz (approved):
@genryxy thanks, merging

0crat:
Job `gh:artipie/artipie#827` is not assigned, can't get performer

Issue #847, opened 2021-02-03
Examples scripts


g4s8:
Let's improve examples scripts

olenagerasimova:
@g4s8 after this improvements smoke-test action is always successful, even if some proof actually failed. For example
https://github.com/artipie/artipie/pull/857/checks?check_run_id=1987434442 
maven prof failed here.

PR #848, opened 2021-02-03
Examples scripts

g4s8:
#847 - added common bash utils for all examples, refactored npm to use it, added `trap` to stop artipie container in any case

### DIFF
--- a/examples/npm/run.sh
+++ b/examples/npm/run.sh
@@ -1,25 +1,12 @@
-set -x
-set -e
+basedir="$(dirname $(readlink -f $0))"
+source "${basedir}/../utils.sh"
 
 image="$1"
-if [[ -z "$image" ]]; then
-  image="artipie/artipie:latest"
-fi
-basedir="$(dirname $(readlink -f $0))"
+port=8080
 
-# Start artipie.
-docker run --rm --detach --name artipie \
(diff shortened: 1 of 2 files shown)


olenagerasimova (approved):
@g4s8 thanks, looks good. But, personally, I'd prefer the way the scripts were - when entire step by step process is duplicated in each example. It's simpler to read and faster to understand single file script than several files script.

g4s8:
> @g4s8 thanks, looks good. But, personally, I'd prefer the way the scripts were - when entire step by step process is duplicated in each example. It's simpler to read and faster to understand single file script than several files script.

@olenagerasimova there were a bug in a script related to docker lifecycle: if the script fails in the middle the docker container will never be stopped. So now we have to fix this bug in each script, this is why I created a function to start docker and automatically stop it on script finish or error. If we'll have a bug or improvement in future for docker starting, we'll change only base script, and not each of the examples.

olenagerasimova:
@g4s8 sure the way you suggested is easier to fix and maintain, no arguments here. I've expressed my opinion from external user point of view, but I guess it's the matter of taste.

0crat:
Job `gh:artipie/artipie#848` is not assigned, can't get performer

Issue #855, opened 2021-02-18
[pdd] Fix integration tests: use Docker image via testcontainers on verify phase


0pdd:
The puzzle `851-a9da7c9e` from #851 has to be resolved: 

https://github.com/artipie/artipie/blob/a492c3d2cbf453d3829e3d1125f44ec15e7ce29d/.github/workflows/integration-tests.yml#L29-L33

The puzzle was created by @olenagerasimova on 29-Jan-21. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@0pdd 4 puzzles [#860](https://github.com/artipie/artipie/issues/860), [#861](https://github.com/artipie/artipie/issues/861), [#862](https://github.com/artipie/artipie/issues/862), [#863](https://github.com/artipie/artipie/issues/863) are still not solved.

g4s8:
@zuoqin can you please take this issue and implement integration test for pypi-adapter?

0pdd:
@0pdd 3 puzzles [#861](https://github.com/artipie/artipie/issues/861), [#862](https://github.com/artipie/artipie/issues/862), [#863](https://github.com/artipie/artipie/issues/863) are still not solved; solved: [#860](https://github.com/artipie/artipie/issues/860).

0pdd:
@0pdd 2 puzzles [#861](https://github.com/artipie/artipie/issues/861), [#863](https://github.com/artipie/artipie/issues/863) are still not solved; solved: [#860](https://github.com/artipie/artipie/issues/860), [#862](https://github.com/artipie/artipie/issues/862).

0pdd:
The puzzle `851-a9da7c9e` has disappeared from the source code, that's why I closed this issue.

0pdd:
@0pdd 2 puzzles [#861](https://github.com/artipie/artipie/issues/861), [#977](https://github.com/artipie/artipie/issues/977) are still not solved; solved: [#860](https://github.com/artipie/artipie/issues/860), [#862](https://github.com/artipie/artipie/issues/862), [#863](https://github.com/artipie/artipie/issues/863).

[...]

PR #874, opened 2021-03-18
Removed debian IT with auth

olenagerasimova:
Part of #855 
Removed debian IT with authorization and permissions, also removed unused deb resources. 
We discussed with Kirill that such ITs are not required as we have auth tests in each adapter, tests for permissions in artipie and tests for basic auth in http.


g4s8 (approved):

0crat:
Job `gh:artipie/artipie#874` is not assigned, can't get performer

PR #876, opened 2021-03-25
IT for rpm with TestDeployment

olenagerasimova:
Part of #855 
Remade IT for rpm to use `TestDeployment`

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -26,6 +26,8 @@ jobs:
           key: ${{ runner.os }}-jdk-${{ matrix.java }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-jdk-${{ matrix.java }}-maven-
+      - name: Build it with Maven
+          run: mvn -B test -Dtest=DebianITCase,DebianGpgITCase,BinaryITCase,RpmITCase
       # @todo #851:30min Fix integration tests. Integration tests should start Artipie docker
       #  image built on package state (artipie/artipie:1.0-SNAPSHOT) using testcontainers,
       #  and test it against different use cases. Update integration tests and enable
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -26,7 +26,7 @@ jobs:
           key: ${{ runner.os }}-jdk-${{ matrix.java }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-jdk-${{ matrix.java }}-maven-
-      - run: mvn -B verify -Dit.test=DebianITCase,DebianGpgITCase,BinaryITCase,RpmITCase -Pitcase
+      - run: mvn -B verify -Pitcase
       # @todo #851:30min Fix integration tests. Integration tests should start Artipie docker
       #  image built on package state (artipie/artipie:1.0-SNAPSHOT) using testcontainers,
       #  and test it against different use cases. Update integration tests and enable
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -26,7 +26,7 @@ jobs:
           key: ${{ runner.os }}-jdk-${{ matrix.java }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-jdk-${{ matrix.java }}-maven-
-      - run: mvn -B verify -Dtest=DebianITCase,DebianGpgITCase,BinaryITCase,RpmITCase
+      - run: mvn -B verify -Dit.test=DebianITCase,DebianGpgITCase,BinaryITCase,RpmITCase -Pitcase
       # @todo #851:30min Fix integration tests. Integration tests should start Artipie docker
       #  image built on package state (artipie/artipie:1.0-SNAPSHOT) using testcontainers,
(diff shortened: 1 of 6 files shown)


src/test/java/com/artipie/rpm/RpmITCase.java
```diff
@@ -23,201 +23,77 @@
...
+            "--upload-file", "/w/time-1.7-45.el7.x86_64.rpm"
+        );
+        this.containers.assertExec(
+            "Failed to install time package",
+            new MavenITCase.ContainerResultMatcher(
+                new IsEqual<>(0),
+                new StringContainsInOrder(new ListOf<>("time-1.7-45.el7.x86_64", "Complete!"))
+            ),
+            "yum", "-y", "repo-pkgs", "example", "install"
+        );
```
genryxy, line 96: @olenagerasimova in this method also installation is checked, therefore probably to rename this method a bit

genryxy (changes requested):
@olenagerasimova thanks. Do I understand correctly that verifying `list` operation is not necessary in Artipie module because it is tested in `rpm-adapter`?

olenagerasimova:
@genryxy thanks, corrected

genryxy:
@olenagerasimova Do I understand correctly that verifying `list` operation is not necessary in Artipie module because it is tested in rpm-adapter?

olenagerasimova:
@genryxy yes, same goes to authorisation.

genryxy (commented):
@olenagerasimova thanks, looks good to me
@g4s8 please check/merge

genryxy (dismissed):

.github/workflows/integration-tests.yml
```diff
@@ -26,6 +26,8 @@ jobs:
           key: ${{ runner.os }}-jdk-${{ matrix.java }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-jdk-${{ matrix.java }}-maven-
+      - run: mvn -B package -Pdocker-build
+      - run: mvn -B test -Dtest=DebianITCase,DebianGpgITCase,BinaryITCase,RpmITCase
```
g4s8, line 30: @olenagerasimova integration tests are started by failsafe plugin in `verify` phase, not `test`, and `verify` test depends on `package` phase, so you don't need to call `package` explicitly: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html
Also, `docker-build` profile is conditionally activated if Docker socket file is available (`/var/run/docker.sock`), so you can omit enabling profile instruction.

g4s8 (changes requested):
@olenagerasimova please check my comment

.github/workflows/integration-tests.yml
olenagerasimova: @g4s8 hm, tests are executed before docker image is build, so it does not work. Also, these commands are temporary, we should rewrite the script when all the tests will be migrated to `TestDeployment`, thus I do not see any point in making it perfectly correct right now

g4s8: @olenagerasimova docker image is built on `package` phase but integration tests are started on `verify` (depends on `package`). The lifecicle is `validate` -> `compile` -> `test` (only unit tests with surefire) -> `package` (creates docker image) -> `verify` (failsafe integration tests). Failsafe plugin is running only on `verify` phase, not on `test`, so it doesn't make sense to specify `*ITCase` classes for `test` phase. All unit tests doesn't depend on docker image, so it's safe to run it before `package` phase. Current config: `mvn -B test -Dtest=DebianITCase,DebianGpgITCase,BinaryITCase,RpmITCase` doesn't start any integration tests, only unit tests

olenagerasimova:
@g4s8 updated

.github/workflows/integration-tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           key: ${{ runner.os }}-jdk-${{ matrix.java }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-jdk-${{ matrix.java }}-maven-
-      - run: mvn -B verify -Dtest=DebianITCase,DebianGpgITCase,BinaryITCase,RpmITCase
+      - run: mvn -B verify -Dit.test=DebianITCase,DebianGpgITCase,BinaryITCase,RpmITCase -Pitcase
```
g4s8, line 29: @olenagerasimova why do we need to specify integration tests to run? Failsafe plugin runs all tests ends with `ITcase` by default.

g4s8 (changes requested):
@olenagerasimova thanks, please see my question

.github/workflows/integration-tests.yml
olenagerasimova: @g4s8 same reason you have commented running ITs in this workflow at the beginning of the issue this PR is addressed to: we are remaking ITs and not all them are remade yet, some tests in the "old format" are failing.

g4s8: @olenagerasimova I see. In that case it could be better to disable the failing test with `@Disabled` annotation - it'll be easier to understand what tests needs to be fixed and the developer won't forget to add the test to CI configuration when fix it. WDYT?

olenagerasimova:
@g4s8 disabled tests

g4s8 (approved):
@olenagerasimova thanks

olenagerasimova:
@g4s8 please, merge

PR #881, opened 2021-04-26
Maven proxy test with TestDeployment

olenagerasimova:
Part of #855 
Remade `MavenProxyIT` to use `TestDeployment`.


genryxy (approved):
@olenagerasimova thanks, looks good

PR #883, opened 2021-04-27
Extended TestDeployment to work with several artipie containers

olenagerasimova:
Part of #855 
I've extended `TestDeployment` to be able to work with several artipie containers. Also remade `PypiProxyITCase` but had to disable it due to the problem with [urls](https://github.com/artipie/pypi-adapter/issues/153).


olenagerasimova:
@g4s8 please, take a look, I need these changes to remade other proxy tests

g4s8 (approved):
@olenagerasimova looks good, thanks

PR #885, opened 2021-04-28
Files proxy IT with `TestDeployment`

olenagerasimova:
Part of #855 
Remade `FileProxyAuthIT` to use `TestDeployment`, moved and renamed `BinaryITCase`, moved resources to separate directory.


g4s8 (changes requested):
@olenagerasimova maybe we can place all integration tests in one package? since it's actually not running against some specific class or package but against Artipie server

olenagerasimova:
@g4s8 we could, but I'm not sure if it will be convenient: there are several test classes for each adapter, so the common package will have a lot of classes, and corresponding package in test resources with repository settings also will very huge.

g4s8:
> @g4s8 we could, but I'm not sure if it will be convenient: there are several test classes for each adapter, so the common package will have a lot of classes, and corresponding package in test resources with repository settings also will very huge.

@olenagerasimova but in that case it'll be not clear where to find integration test class for adapters, e.g. we don't have package `com.artipie.rpm` in `com.artipie:artipie` module, and the test is not exactly is running against `rpm` package but for Artipie web service with rpm module enabled

olenagerasimova:
@g4s8 hm, it's clear enough from my point of view. If we want it transparent, than package with these tests should be called `adapters integration tests`

g4s8 (approved):
@olenagerasimova ok, let's keep it here

PR #887, opened 2021-04-30
`NpmProxyITCase` with `TestDeployment`

olenagerasimova:
Part of #855 
Remade `NpmProxyITCase` use `TestDeployment`.


g4s8 (approved):
Thanks

PR #890, opened 2021-05-11
WIP: Issue #889 Adding Monitoring Metrics

### LINKED ISSUE #889: Monitoring and metrics tool
Add monitoring support for Artipie. Now all metrics are collected into the log, let's add additional metrics provider, like https://github.com/prometheus/prometheus

milkysoft:
close #889


src/main/java/com/artipie/metrics/prometheus/PushGateway.java
```diff
@@ -0,0 +1,301 @@
...
+ * See <a href="https://github.com/prometheus/pushgateway">https://github.com/prometheus/pushgateway</a>
+ */
+public class PushGateway {
+
+    private static final int MILLISECONDS_PER_SECOND = 1000;
+
+    // Visible for testing.
+    protected final String gatewayBaseURL;
+
+    private HttpConnectionFactory connectionFactory = new DefaultHttpConnectionFactory();
```
olenagerasimova, line 66: @milkysoft looks like this should be a final field initialized in the ctor, also compound names are not allowed by our checkstyle. Please, read this [blog post](https://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html).

src/main/java/com/artipie/metrics/prometheus/PrometheusGauge.java
```diff
@@ -0,0 +1,54 @@
...
+package com.artipie.metrics.prometheus;
+
+import com.artipie.metrics.Gauge;
+
+/**
+ * {@link Gauge} implementation storing data in memory.
+ *
+ * @since 0.8
+ */
+public final class PrometheusGauge implements Gauge {
```
olenagerasimova, line 33: @milkysoft I'd suggest to come up with some shortening for `Prometheus` and add it as prefix for the class names, smth like `PrmthGuage`, `PrmthCount`, etc.

olenagerasimova (changes requested):
@milkysoft please, check my comments. Also, there are following suggestions for the future:
1. This PR is too large, there should be about 100-200 changes lines per PR, so it would be much better to decompose your work: 
a) create `HttpConnection` interface and implementation, create `PrometheusCounter` and unite test for it
b) create gauge and metrics and corresponding unit tests
c) create `PushGateway` and unit test
Thus, there will be 3 PR (which is better for your performance metrics), these PRs will be reviewed and merged faster as they are smaller, code will be unit tested. 
2. Please, do not submit a PR if qulice is complaining about anything. You should get used to these strict rules and write code in the style we use here, otherwise you will spend days fixing checkstyle.
3. Read Yegor's blog posts about java code and his understanding of the OOP. I know it's very unusual and frustrated at first, but in artipie we follow this style.
4. Please, do not add any more code in this PR. Instead, correct this PR description not to close connected issue (change word 'close' to 'part of') and add unit tests in the next PR.

I will not review this PR until all [these](https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md#pull-request-style) rules are followed and checksyle checks passes.

Issue #352, opened 2020-07-17
Add footer to Artipie central


olenagerasimova:
Let's add footer to Artipie central, the footer should contain at least currently deployed version, links to github repository and telegram chat. Also it would be nice to see amount of users, repositories and uploaded artifacts.

g4s8:
I think we need to start with a version details here.

genryxy:
Displaying of used version of Artipie after starting server was implemented. Logs look like:
```
$ docker run --name artipie -v C:/work/w_play_artipie_in_docker/artipie3:/var/artipie -p 8080:8080 --rm --user=artipie:artipie artipie/artipie:1.0-SNAPSHOT
[INFO] 2022-01-24 08:46:09,480 main com.artipie.VertxMain - Used version of Artipie: 1.0-SNAPSHOT
....
```

dgarus:
@olenagerasimova @g4s8 
Could we close this issue?

olenagerasimova:
@dgarus we have not added footer dashboard... If you think we do not need to, you can close

dgarus:
@olenagerasimova ok. Let's pay in account this issue when we will rebuild the front.

PR #911, opened 2021-06-10
Add version endpoint and footer about version for image

### LINKED ISSUE #526: Version endpoint
Currently, it's not possible to get release version of Artipie web app. Let's add new endpoint `/.version` to show current version from `pom.xml`. (it can be done using `filtering` option of Maven resources plugin to propagate any build value to properties file and retrieve programmatically). Don't forget to check if Docker build action sets the version correctly.

genryxy:
Closes #526 
Part of #352 
Added endpoint for obtaining version.
Also added info about version to built image.


olenagerasimova (dismissed):
@genryxy thanks, looks good to me

genryxy:

g4s8 (changes requested):
@genryxy did you try to build and run this docker? It doesn't work for me:
```
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Exception in thread "main" java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Unknown Source)
	at com.artipie.JavaResource.copy(JavaResource.java:63)
	at com.artipie.VertxMain.settings(VertxMain.java:157)
	at com.artipie.VertxMain.start(VertxMain.java:88)
	at com.artipie.VertxMain.main(VertxMain.java:143)
```

genryxy:
@g4s8 could you please take a look

g4s8 (changes requested):
@genryxy two questions here:
 1. Why did you move all existing resources to a new resource directory? (e.g. `src/main/resources/example` to `src/main/resources-filtered/example/`
 2. This PR introduces two methods to access the version: first version is passed via Docker build argument parameter, second is filtered from properties file. I think we can use only one method to access the version, no need to use both at the same time

genryxy:
@g4s8 thanks left only one method for obtaining version. As for existing resources to a new directory, otherwise it does not start Artipie on localhost because files are copied in `VertxMain` and they can't be found

genryxy:
@g4s8 could you please take a look

src/main/java/com/artipie/http/VersionSlice.java
```diff
@@ -0,0 +1,55 @@
...
+        final Publisher<ByteBuffer> body
+    ) {
+        return new AsyncResponse(
+            CompletableFuture.supplyAsync(
+                () -> new RsWithStatus(
+                    new RsJson(
+                        Json.createArrayBuilder().add(
+                            Json.createObjectBuilder().add(
+                                "version",
+                                new ArtipieProperties().version()
```
g4s8, line 46: @genryxy I think it's better to load properties outside this class and pass it as a `Properties` class state via constructor, then we'll easily  test HTTP layer by passing test customized properties object.

g4s8 (changes requested):
@genryxy please check my comment. Also, there are multiple files renamed without need:
```
0 src/main/resources/example/artipie.yaml → ...n/resources-filtered/example/artipie.yaml
0 .../resources/example/repo/_credentials.yaml → ...s-filtered/example/repo/_credentials.yaml
0 .../resources/example/repo/_permissions.yaml → ...s-filtered/example/repo/_permissions.yaml
0 ...ain/resources/example/repo/_storages.yaml → ...rces-filtered/example/repo/_storages.yaml
0 src/main/resources/log4j.properties → src/main/resources-filtered/log4j.properties 
```

genryxy:
@g4s8 thanks, otherwise these files are not found if Artipie is started from a new folder, therefore these files were moved to `resources-filtered` folder. It happened after adding to pom
```
<resources>
  <resource>
    <directory>src/main/resources-filtered</directory>
    <filtering>true</filtering>
  </resource>
</resources>
```
Now I removed this addition as it works without these lines. Also, passed properties via constructor

g4s8:
> <resources>
>   <resource>
>     <directory>src/main/resources-filtered</directory>
>     <filtering>true</filtering>
>   </resource>
> </resources>

@genryxy you can use:
```xml
<resources>
  <resource>
    <directory>src/main/resources-filtered</directory>
    <filtering>true</filtering>
  </resource>
  <resource>
    <directory>src/main/resources</directory>
    <filtering>false</filtering>
  </resource>
</resources>

```

src/main/java/com/artipie/http/VersionSlice.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public Response response(
+        final String line,
+        final Iterable<Map.Entry<String, String>> headers,
+        final Publisher<ByteBuffer> body
+    ) {
+        return new AsyncResponse(
+            CompletableFuture.supplyAsync(
```
g4s8, line 43: Here you can just return `RsJson`, no need to run it async

src/test/java/com/artipie/http/VersionSliceTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Tests for {@link VersionSlice}.
+ * @since 0.21
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ */
+final class VersionSliceTest {
+    @Test
+    void returnVersionOfApplication() throws IOException {
+        final Properties properties = new Properties();
```
g4s8, line 30: The idea of moving `Properties` object to constructor of `VersionSlice` was to separate tests of HTTP object and properties-loading object. In test for `VersionSlice` we can construct `Properties` file without loading it from resource, just creating on the fly:
```java
var prop = new Properties();
prop.put("artipie.version", "1.0-test");
MatcherAssert.assert(new VersionSlice(prop), ...);
```

src/test/java/com/artipie/http/VersionSliceTest.java
```diff
@@ -0,0 +1,49 @@
...
+        properties.load(new TestResource("artipie.properties").asInputStream());
+        MatcherAssert.assertThat(
+            new VersionSlice(new ArtipieProperties()),
+            new SliceHasResponse(
+                Matchers.allOf(
+                    new RsHasStatus(RsStatus.OK),
+                    new RsHasBody(
+                        String.format(
+                            "[{\"version\":\"%s\"}]",
```
g4s8, line 39: It's not always valid to verify json with a string (since it can be valid in multiple formats, e.g. with white-spaces, newlines, etc). So here [json matcher for body](https://github.com/artipie/http/blob/master/src/main/java/com/artipie/http/hm/IsJson.java) could be used: `new RsHasBody(new IsJson(new JsonHas("version", new JsonValueIs("1.0")))`

g4s8 (changes requested):
@genryxy please check a few comments

genryxy:
@g4s8 thanks, corrected with response and matcher. As for properties, `VersionSlice` accept `ArtipieProperties` which does not extend `Properties` class, which can be manually filled. If `VersionSlice` will accept `Properties` instead of `ArtipieProperties` it would be necessary to specify key of values which should be obtained.

g4s8 (approved):
@genryxy thanks

Issue #869, opened 2021-03-17
Documentation imrpovements


g4s8:
Since we're getting more people to the project, let's pay attention to contributing guides:
 - CONTRIBUTING.md file with guides
 - README.md with documentation
 - PR and issues templates

PR #915, opened 2021-06-18
Described file and file-proxy repo

g4s8:
For #869 - added `file` and `file-proxy` descriptions with examples.


olenagerasimova (approved):
@g4s8 thanks

Issue #903, opened 2021-06-03
Failed to upload big file1G  to central.artipie.com via files-adapter


g4s8:
Reproduce:
 - Create binary repository on Central, configure permissions and users
 - Upload 1G file to this repo

Expected behavior: file uploaded successsfully
Actual behavior: `Connection reset by peer` error or `Broken pipe` error

Maybe it could be reproduced on local deployment too. Maybe it depends on timeouts and can be reproduced with long upload of smaller files.

Discovered by @aiovlev

genryxy:
@g4s8 I've managed to reproduce failed upload of big file
Steps to reproduce
1) Start artipie locally
2) Create binary repository
3) Generate big file. For example, 10GB
`dd if=/dev/zero of=big.file bs=4096 count=2621440`
4) Push generated file with specifying max time as 15 minutes
`curl -i -X PUT --data-binary "@big.file" -u artipie:artipie --max-time 900 http://localhost:8080/artipie/bin`

Client: 
```
$ curl -i -X PUT --data-binary "@big.file" -u artipie:artipie --max-time 900 http://localhost:8080/artipie/bin
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 26 10.0G    0    28   26 2675M      0  44.5M  0:03:49  0:01:00  0:02:49 44.1MHTTP/1.1 100 Continue

HTTP/1.1 503 Service Unavailable
transfer-encoding: chunked

Request cancelled by timeout
```

Server:
```
[INFO] 2021-06-04 13:02:09,175 vert.x-eventloop-thread-1 com.artipie.http.MainSlice - >> PUT /artipie/bin HTTP_1_1

Host: localhost:8080
Authorization: Basic YXJ0aXBpZTphcnRpcGll
User-Agent: curl/7.64.0
Accept: */*
Content-Length: 10737418240
Content-Type: application/x-www-form-urlencoded
Expect: 100-continue
[INFO] 2021-06-04 13:02:09,215 rio-4 com.artipie.http.MainSlice - << CONTINUE
[INFO] 2021-06-04 13:02:09,217 rio-4 com.artipie.http.auth.Permissions$$Lambda$300/0x0000000800c15840 - Operation 'w' allowed for '*'
[INFO] 2021-06-04 13:02:43,874 pool-1-thread-1 com.artipie.metrics.Metrics - Counters:
http.response.body.size: 0
[...]

PR #917, opened 2021-06-21
Removed application timeout slice

g4s8:
For #903 - removed connection timeout on application level,
since it should be configured at server or balancer.


genryxy (approved):
@g4s8 thanks, looks good. I've tested locally to upload 10GB and after this fix it works well

PR #922, opened 2021-08-05
Added cache for configuration files

### LINKED ISSUE #195: Repository configuration cache
Artipie performs a lot of read operations for repository configs, it would be better to cache these configs in memory. The config can be updated at any time, so we need to drop cached value after some time, e.g. after few minutes. Consider using soft references to avoid OOM errors when working with many repos.

genryxy:
Closes #195 
Added cache with soft references with eviction which is equal to two minutes for repository configuration files and for aliases with configs of storage.

### DIFF
--- a/src/main/java/com/artipie/ArtipieProperties.java
+++ b/src/main/java/com/artipie/ArtipieProperties.java
@@ -20,6 +20,11 @@ public final class ArtipieProperties {
     public static final String VERSION_KEY = "artipie.version";
 
     /**
+     * Expiration time for cache of configuration files.
+     */
+    static final String CONFIG_TIMEOUT = "artipie.config.cache.timeout";
+
+    /**
      * Name of file with properties.
      */
     private final String filename;
@@ -58,15 +63,8 @@ public final class ArtipieProperties {
      * Obtains timeout of caching configuration files in milliseconds.
      * @return Timeout of caching files with configuration.
      */
-    public int configCacheTimeout() {
-        int time;
-        try {
-            time = Integer.parseInt(this.properties.getProperty("artipie.config.cache.timeout"));
-        } catch (final NumberFormatException exc) {
-            // @checkstyle MagicNumberCheck (1 line)
-            time = 2 * 60 * 1000;
-        }
-        return time;
+    public String configCacheTimeout() {
+        return this.properties.getProperty(ArtipieProperties.CONFIG_TIMEOUT);
     }
 
     /**
--- a/src/main/java/com/artipie/ArtipieProperties.java
+++ b/src/main/java/com/artipie/ArtipieProperties.java
@@ -55,6 +55,21 @@ public final class ArtipieProperties {
     }
 
     /**
+     * Obtains timeout of caching configuration files in milliseconds.
+     * @return Timeout of caching files with configuration.
+     */
+    public int configCacheTimeout() {
+        int time;
+        try {
+            time = Integer.parseInt(this.properties.getProperty("artipie.config.cache.timeout"));
+        } catch (final NumberFormatException exc) {
+            // @checkstyle MagicNumberCheck (1 line)
+            time = 2 * 60 * 1000;
+        }
+        return time;
+    }
+
+    /**
      * Load content of file.
      */
     private void loadProperties() {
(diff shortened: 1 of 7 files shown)


olenagerasimova (approved):
@genryxy thanks, looks good for me

genryxy:
@g4s8 could you please check / merge

src/main/java/com/artipie/RepositoriesFromStorage.java
```diff
@@ -4,19 +4,65 @@
...
+     */
+    private static LoadingCache<KeyAndStorage, Single<StorageAliases>> aliases;
+
+    static {
+        final CacheLoader<KeyAndStorage, Single<String>> ldrconfigs = new CacheLoader<>() {
+            @Override
+            public Single<String> load(final KeyAndStorage pair) {
+                return Single.fromFuture(
+                    new ConfigFile(pair.key)
+                        .valueFrom(pair.storage)
```
g4s8, line 43: I don't like private fields access here because it increases the coupling between components. I'd suggest introducing a method in `KeyAndStorage` which can access these private fields, e.g.
```
// class KeyAndStorage
CompletionStage<Content> content() {
  return new ConfigFile(this.key).valueFrom(this.storage);
}
```
Also, in this case, it could have a more meaningful name, e.g. `ConfigContent` or similar

src/test/java/com/artipie/RepositoriesFromStorageTest.java
```diff
@@ -85,7 +85,7 @@ void throwsExceptionWhenConfigYamlAbsent() {
             this::repoConfig
         );
         MatcherAssert.assertThat(
-            result.getCause(),
+            result.getCause().getCause(),
```
g4s8, line 88: This change looks not related to PR modification, isn't it?

g4s8 (changes requested):
@genryxy looks good, please check a few comments

src/test/java/com/artipie/RepositoriesFromStorageTest.java
genryxy: @g4s8 now `ValueNotFoundException` is wrapped with `java.util.concurrent.ExecutionException` therefore this test was changed

src/main/java/com/artipie/ArtipieProperties.java
```diff
@@ -54,6 +54,21 @@ public String version() {
...
+     * Obtains timeout of caching configuration files in milliseconds.
+     * @return Timeout of caching files with configuration.
+     */
+    public int configCacheTimeout() {
+        int time;
+        try {
+            time = Integer.parseInt(this.properties.getProperty("artipie.config.cache.timeout"));
+        } catch (final NumberFormatException exc) {
+            // @checkstyle MagicNumberCheck (1 line)
+            time = 2 * 60 * 1000;
```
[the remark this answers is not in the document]
genryxy: @g4s8 I've replaced with `Integer.getInteger("artipie.config.cache.timeout", 2 * 60 * 1000)` although it does not throw any exception in case of mistyping in `artipie.properties` and I had to set property value at first

genryxy:
@g4s8 thanks, corrected and answered

src/main/java/com/artipie/RepositoriesFromStorage.java
```diff
@@ -35,33 +35,33 @@
...
+        System.setProperty(
+            ArtipieProperties.CONFIG_TIMEOUT,
+            new ArtipieProperties().configCacheTimeout()
+        );
+        final int timeout = Integer.getInteger(ArtipieProperties.CONFIG_TIMEOUT, 2 * 60 * 1000);
         RepositoriesFromStorage.configs = CacheBuilder.newBuilder()
             .expireAfterWrite(timeout, TimeUnit.MILLISECONDS)
-            .softValues().build(ldrconfigs);
+            .softValues()
+            .build(
```
g4s8, line 46: I mean can we use something like `build(config -> config.configContent())` here? Or `CacheLoader` is not a `FunctionalInterface`?

g4s8 (commented):
@genryxy please check one more questions

src/main/java/com/artipie/RepositoriesFromStorage.java
genryxy: @g4s8 got it. `CacheLoader` is an abstract class.

g4s8 (approved):
@genryxy thanks

Issue #460, opened 2020-08-19
CachedAuth.java:45-48: Refactor this class to move cache...


0pdd:
The puzzle `442-4070cd49` from #442 has to be resolved: 

https://github.com/artipie/artipie/blob/de6708ff7bb634e915e2b64448a4f79d35f10f46/src/main/java/com/artipie/auth/CachedAuth.java#L45-L48

The puzzle was created by Kirill on 18-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #924, opened 2021-08-06
Specify expiration time and refactor `CachedAuth`

### LINKED ISSUE #339: CachedAuth.java:40-44: Specify expiration time...
The puzzle `285-b75b90d5` from #285 has to be resolved: 

https://github.com/artipie/artipie/blob/36d95233470090223c491a89f945f48ad1f55b8c/src/main/java/com/artipie/auth/CachedAuth.java#L40-L44

The puzzle was created by Kirill on 15-Jul-20. 

genryxy:
Closes #339 #460 
Specify expiration time for auth, used guava implementation.
Extracted `AuthCache` interface for creating instance of `CachedAuth` only once. Created `CachedAuthWrap` for using passed instance of `AuthCache` in `YamlSettings#auth` .

### DIFF
--- a/src/main/java/com/artipie/YamlSettings.java
+++ b/src/main/java/com/artipie/YamlSettings.java
@@ -8,7 +8,7 @@ import com.amihaiemil.eoyaml.YamlMapping;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
 import com.artipie.asto.SubStorage;
-import com.artipie.auth.CachedAuth;
+import com.artipie.auth.AuthCache;
 import com.artipie.auth.GithubAuth;
 import com.artipie.http.auth.Authentication;
 import com.artipie.http.slice.KeyFromPath;
@@ -38,11 +38,18 @@ public final class YamlSettings implements Settings {
     private final YamlMapping content;
 
     /**
+     * Authentication cache.
+     */
+    private final AuthCache authcache;
+
+    /**
      * Ctor.
      * @param content YAML file content.
+     * @param authcache Auth cache
      */
-    public YamlSettings(final YamlMapping content) {
+    public YamlSettings(final YamlMapping content, final AuthCache authcache) {
         this.content = content;
+        this.authcache = authcache;
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


src/main/java/com/artipie/YamlSettings.java
```diff
@@ -126,4 +136,35 @@ public Storage repoConfigsStorage() {
     private static boolean hasTypeFile(final YamlMapping cred) {
         return cred != null && "file".equals(cred.string("type"));
     }
+
+    /**
+     * Wrapping for auth cache.
+     * @since 0.22
+     */
+    private static final class CachedAuthWrap implements Authentication {
```
olenagerasimova, line 144: I'd suggested to simplify the name of the class to `Cached` as `Wrap` is usually used for abstract classes, this more looks like cached decorator of `Authentication`

src/main/java/com/artipie/auth/CachedAuth.java
```diff
@@ -18,53 +20,113 @@
...
- *  X minutes, then remove only this token. Consider using Guava's time-evicted-cache
- *  implementation: https://github.com/google/guava/wiki/CachesExplained#eviction
- * @todo #442:30min Refactor this class to move cache field to instance field
- *  and make sure the instance is not created on each request. Right now we're using
- *  a hotfix - CachedAuth created from settings on each request but uses static hash map
- *  field for caching.
  */
-public final class CachedAuth implements Authentication {
-
+public final class CachedAuth implements AuthCache {
```
olenagerasimova, line 24: I'd suggest to think about other name for this class: for example, we can name it by expiration method `ExpiredAfterAccess` or by cache implementation used inside.

olenagerasimova (changes requested):
@genryxy thanks, several suggestions for naming, take a look please

src/main/java/com/artipie/auth/CachedAuth.java
genryxy: @olenagerasimova I don't think that `ExpiredAfterAccess` is more suitable because it is possible that internal realization will change. For example `expireAfterWrite` or something else. Renamed to `CachedUsers`. Hope it will be a bit clearer

genryxy:
@olenagerasimova thanks, answered and renamed

olenagerasimova (approved):
@genryxy thanks
@g4s8 please, take a look/merge

g4s8 (approved):
@genryxy thanks

PR #927, opened 2021-08-16
Add test for authentication without credentials section

### LINKED ISSUE #299: The credentials section should be optional
The credential section in the artipie.yaml file should be optional by default:
```yaml
meta:
  storage:
    type: fs
    path: ./configs
  credentials:
    type: file
    path: _credentials.yaml
  layout: org
  metrics:
    type: log
```

Anyone should be able to do whatever possible by default

genryxy:
Closes #299 
Added test which verifies that user is authenticated when `credentials` section is absent


g4s8 (approved):
@genryxy thanks

PR #928, opened 2021-08-19
Removed not actual @todo

### LINKED ISSUE #470: ArtipieServer.java:83-89: Extract class for building...

olenagerasimova:
Closes #470 
I do not think refactoring from this `todo` is actually required: as there are basically only two places in test code where `artipie.yaml` is created - `ArtipieServer` and test for settings `YamlSettingsTest`. `YamlSettingsTest` has a lot of different variants of artipie.yaml (to test that in all cases settings are read correctly) and for test it's ok (and even better in my opinion) to have all these variants as string stabs.

### DIFF
--- a/src/test/java/com/artipie/ArtipieServer.java
+++ b/src/test/java/com/artipie/ArtipieServer.java
@@ -186,13 +186,6 @@ public class ArtipieServer {
      *
      * @return Port the servers listening on.
      * @throws IOException In case of error creating configs or running the server.
-     * @todo #449:30min Extract class for building settings in YAML format.
-     *  Building of settings YAML for usage in tests occurs more the once.
-     *  It is used to build setting in `ArtipieServer`
-     *  and in `YamlSettingsTest` to create settings examples in unit tests.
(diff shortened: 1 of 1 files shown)


genryxy:
@olenagerasimova thanks, I think that class for creating config file could be useful in some cases. Something similar to [`RepoConfigYaml`](https://github.com/artipie/artipie/blob/ee3f98590619112f60b81be7c8751b9d623180eb/src/test/java/com/artipie/RepoConfigYaml.java) which you created earlier.

olenagerasimova:
@genryxy I cannot argue with 

> class for creating config file could be useful in some cases

But I've looked through the test code and do not see this "some cases" except for `ArtipieServer`...

genryxy (approved):

Issue #933, opened 2021-09-02
Integrate conda-adapter


olenagerasimova:
Let's integrate conda-adapter  to artipie ecosystem:
- add corresponding dependency and `CondaSlice` to `SliceFromConfig`
- create integration test for install and upload
- add description to documentation
- add smoke test
- add repository to artipie dashboard

When done, do not forget to release artipie

PR #934, opened 2021-09-02
Added conda-adapter dependency and IT for install

olenagerasimova:
Part of #933 
Added conda-adapter dependency, slice to `SliceFromConfig` and IT for install


src/test/java/com/artipie/conda/CondaITCase.java
```diff
@@ -0,0 +1,96 @@
...
+            new ContainerResultMatcher(
+                new IsEqual<>(0),
+                new StringContainsInOrder(
+                    new ListOf<String>(
+                        "The following packages will be downloaded:",
+                        "http://artipie:8080/my-conda",
+                        "linux-64::snappy-1.1.3-0",
+                        "Preparing transaction: ...working... done",
+                        "Verifying transaction: ...working... done",
+                        "Executing transaction: ...working... done"
```
g4s8, line 82: I don't think checking output logs is a good idea here, it could change with new version of conda. Checking exit code would be enough here.

g4s8 (changes requested):
@olenagerasimova thanks, please check my comments

src/test/java/com/artipie/conda/CondaITCase.java
olenagerasimova: @g4s8 I don't agree that checking exit code is ever enough: for example, exit code will always be ok if we install package from central (global) repository, not from artipie.

olenagerasimova:
@g4s8 please check my comment, also I've shortened output logs check

olenagerasimova:
@g4s8 corrected as we discussed on the meet up, please, take a look/merge

g4s8 (approved):
@olenagerasimova thanks

olenagerasimova:
@g4s8 please, approve again, pdd check was failing, I've fixed its config

g4s8 (approved):

PR #937, opened 2021-09-06
Anaconda example

olenagerasimova:
Part of #933 
Added anaconda docker example


examples/conda/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM continuumio/miniconda3:4.10.3
+
+WORKDIR /test
+RUN conda install -y conda-build
+RUN conda install -y conda-verify
+RUN conda install -y anaconda-client
```
g4s8, line 6: Can we compact it into one command?
```suggestion
RUN conda install -y conda-build conda-verify anaconda-client
```

examples/conda/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN anaconda config --set url "http://artipie:8080/my-conda/" -s
+RUN conda config --set anaconda_upload yes
+RUN echo "channels:\r\n  - http://artipie:8080/my-conda" > /root/.condarc
```
g4s8, line 9: And here (as fewer `RUN` command we have, as faster Docker builds images; each `RUN` produces image layer which is not really fast operation, see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/):
 ```suggestion
RUN anaconda config --set url "http://artipie:8080/my-conda/" -s && \
    conda config --set anaconda_upload yes && \
    echo "channels:\r\n  - http://artipie:8080/my-conda" > /root/.condarc
```

**Review by g4s8 — CHANGES_REQUESTED** [review]
@olenagerasimova please check my suggestions

**Comment by olenagerasimova** [comment]
@g4s8 thanks, I've corrected your comments, but I have a problem with test project directory:
```
Step 5/7 : ADD sample-project /test/sample-project
ADD failed: file not found in build context or excluded by .dockerignore: stat sample-project: file does not exist
+ docker run --name=smoke-conda --rm --net=artipie -v /var/run/docker.sock:/var/run/docker.sock test/conda
+ tee -a /home/runner/work/artipie/artipie/examples/out.log
Unable to find image 'test/conda:latest' locally
docker: Error response from daemon: pull access denied for test/conda, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
```
I've also tried `copy ./sample-project /test/sample-project`, it fails with the same error.

g4s8:
> @g4s8 thanks, I've corrected your comments, but I have a problem with test project directory:
> 
> ```
> Step 5/7 : ADD sample-project /test/sample-project
> ADD failed: file not found in build context or excluded by .dockerignore: stat sample-project: file does not exist
> + docker run --name=smoke-conda --rm --net=artipie -v /var/run/docker.sock:/var/run/docker.sock test/conda
> + tee -a /home/runner/work/artipie/artipie/examples/out.log
> Unable to find image 'test/conda:latest' locally
> docker: Error response from daemon: pull access denied for test/conda, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
> See 'docker run --help'.
> ```
> 
> I've also tried `copy ./sample-project /test/sample-project`, it fails with the same error.

@olenagerasimova first: try using `COPY` instead of magic `ADD` command - `ADD` using URIs for `<src>` params, where `COPY` uses local files, see https://stackoverflow.com/a/24958548/1723695 ;
Second: I don't see `sample-project` directory to `ADD` in your branch, so there are nothing to add here.

Issue #946, opened 2021-10-01
Update and set-up conda adapter


olenagerasimova:
Let's update anaconda adapter and set it up, we need to add settings for:
- authentication token time to leave
- time to clean up expired token (maybe as cron expression)

Then, these setting should be used to create conda slice and set up `quartz` job to clean expired tokens.

olenagerasimova:
done

PR #948, opened 2021-10-04
Added quartz scheduler and conda job

olenagerasimova:
Part of #946 
Added `quartz` dependency, scheduler and job for anaconda. Also, updated `http`, `asto` and `conda` adapter.

### DIFF
--- a/src/main/java/com/artipie/QuartzScheduler.java
+++ b/src/main/java/com/artipie/QuartzScheduler.java
@@ -4,8 +4,7 @@
  */
 package com.artipie;
 
-import com.artipie.asto.Storage;
-import com.artipie.conda.CondaQuartz;
+import com.artipie.repo.QuartzRepoJob;
 import com.artipie.rpm.misc.UncheckedConsumer;
 import com.jcabi.log.Logger;
 import java.util.Collection;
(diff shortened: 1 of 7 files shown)


olenagerasimova:
@g4s8 please, take a look

src/main/java/com/artipie/QuartzScheduler.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public QuartzScheduler(final Storage asto, final Collection<RepoConfig> config) {
+        this.asto = asto;
+        this.config = config;
+    }
+
+    /**
+     * Start scheduler, add shutdown hook and schedule the jobs.
+     */
+    void setup() {
```
g4s8, line 45: Maybe `start` is a better name for this method? (because it actually starts scheduler)

src/main/java/com/artipie/QuartzScheduler.java
```diff
@@ -0,0 +1,72 @@
...
+                        }
+                    }
+                )
+            );
+            scheduler.start();
+            this.config.stream().filter(repo -> repo.type().equals("conda"))
+                .forEach(
+                    new UncheckedConsumer<>(
+                        conda -> new CondaQuartz(this.asto, conda).triggerJob(scheduler)
+                    )
```
g4s8, line 65: Maybe we can provide some common interface from config, e.g. `this.config.stream().flatMap(cfg -> cfg.jobs()).forEach(job -> job.triggerJob(scheduler))`, where `jobs()` can return a collection of jobs to run on quartz scheduler or empty collection if no jobs specified. WDYT?

g4s8 (changes requested):
@olenagerasimova please check my comments

src/main/java/com/artipie/QuartzScheduler.java
olenagerasimova: I do not like the idea add new methods about jobs to `RepoConfig` as not each repo needs any jobs. I've added enum with repositories filter and method to schedule the jobs, thus we can add new items there for any repository or even the same job for several repositories.

olenagerasimova:
@g4s8 thanks, corrected and answered

src/main/java/com/artipie/QuartzScheduler.java
g4s8: @olenagerasimova I see. But what if need to configure these jobs later? E.g. configure by setting intervals or some additional parameters. Then we need to pass it via repository configuration somehow. How we can pass it to enum values?

olenagerasimova: All additional parameters will be added to `settings` section of the repository configuration `RepoConfig`, an instance of the `RepoConfig` is passed to enum `schedule()` method, where all the job trigger settings can be performed. 

It's as with conda now: cron value is set in the repo config `settings` section and passed over `RepoConfig` instance to enum `schedule()` method and to `CondaQuartz` class.

g4s8: Got it, thanks

g4s8 (approved):

PR #949, opened 2021-10-04
ci: bump qulice and enabled for CI pipeline

g4s8:
Updated qulice to `0.19.0` with fix yegor256/qulice#1095
Enabled it in CI pipeline.


olenagerasimova (approved):
@g4s8 thanks

PR #950, opened 2021-10-05
Used Artipie docker image in `MavenMultiProxyIT`

olenagerasimova:
Remade `MavenMultiProxyIT` to use Artipie docker images instead of `ArtipieServer`.


g4s8 (approved):
@olenagerasimova thanks

PR #952, opened 2021-10-06
Fixed several smoke tests

olenagerasimova:
Fixed `smoke-tests.yml` github workflow, go and debian smoke tests.

PR #953, opened 2021-10-06
Remade smoke test for nuget

olenagerasimova:


examples/run.sh
```diff
@@ -165,7 +165,7 @@ fi
 log_debug "tests: ${tests[@]}"
 
 # FIXME: some repository tests don't work, fix them:
-#     binary debian go nuget binary debian go nuget
+# conda
```
g4s8, line 168: @olenagerasimova If this fix is only for Nuget, then we can merge it because Nuget test passed successfully. In that case please put back `go`, `debian` and `conda` for `FIXME` line.
```
test debian - FAILED
test go - FAILED
test conda - FAILED
```

olenagerasimova: @g4s8 debian and go are fixed in #952 , so when #952 is merged (there will be conflicts, I'll resolve them manually) this will be correct.

Issue #87, opened 2020-04-21
Unnecessary operations on each request


HDouss:
On every request received by artipie (when using the default Pie slice), we do the following operations:
* retrieve storage from settings, which reads the yaml content (does not read the file again, but the file content as read at the first time) for "meta -> storage" entry.
* get the value of this storage scalar.

This is corresponding to that code:
```java
CompletableFuture.supplyAsync(
    () -> new Unchecked<>(this.settings::storage).value()
)
```

These two operations are redundant because they will always have the same result, there is no need to make it at every received request.
Ideally, if we won't support hot changing repos config files, we could implement a sort of cache that will return the whole slice (that is decorated by the AsyncSlice) based on the repo. Every repo is associated to a single slice, there is no need to recalculate that at every request. By "the whole slice that is decorated by the AsyncSlice", I meant this part of the code:
```java
CompletableFuture.supplyAsync(
    () -> new Unchecked<>(this.settings::storage).value()
)
.thenComposeAsync(
    storage -> storage.value(new Key.From(String.format("%s.yaml", repo)))
)
.thenApply(content -> new RepoConfig(content))
.thenCompose(Pie::sliceForConfig)

```

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8) please, pay attention to this issue

g4s8:
@HDouss actually we want to support configuration changes, e.g. if you add new repository config, the server should handle this repository on next request. But it would be a good idea to cache configuration for some time, e.g. 5 minutes

HDouss:
@g4s8 Even if we want that, there are redundant operations (even if they are small), retrieving storage from the settings reparses the yaml stored content.

g4s8:
@HDouss but it may be changed, what if you decide to move repository storage from local file system to S3 storage? Then config will be updated and should be parsed again

HDouss:
@g4s8 these operations I am talking about are not about a single repository config but about artipie config, the one with meta. These are not hot-changeable because the file content is stored in `YamlSettings` as a string, but the parsing happens in every request (see `YamlSettings.storage` method)

g4s8:
@HDouss got it, then it's a valid improvement

0crat:
[...]

PR #962, opened 2021-10-28
feat: use `CachedStorages`

genryxy:
Part of #87 
Used `CachedStorages` in `YamlSettings`.
I am not sure about passing `Settings` to `CachedStorages#storage(Settings)` because in `Settings` `CachedStorages` is used. Probably it would be better to pass `YamlMapping` because it is necessary to use only one section from the whole setting but in this case we would depend more on yaml format


g4s8 (approved):
@genryxy thanks, let's merge it. I think we may need to refactor the whole settings/configuration classes later, because now it looks a bit complex @artipie/members WDYT?

genryxy:
@g4s8 probably adding some illustrative examples with expected behaviour would make it a little easier to work with settings and configurations for users

PR #963, opened 2021-11-03
refactor: extract cache package and logic of reading property value

genryxy:
Part of #87 

- Extracted `cache` package.
- Extracted logic of reading value of the property from environment and Artipie properties file.
- Fixed reading value of property and added tests for it. Now it takes value of property from environment, if it was set, otherwise it looks up in `artipie.properties` file. If both of them are `null` it parses to long passed default value.


src/main/java/com/artipie/misc/Property.java
```diff
@@ -0,0 +1,53 @@
...
+     * Ctor.
+     * @param name Name of the property.
+     */
+    public Property(final String name) {
+        this.name = name;
+    }
+
+    /**
+     * Obtains long value of the property set properties or file
+     * with values of the properties.
```
olenagerasimova, line 31: This sentence seems off, please, check

src/main/java/com/artipie/misc/Property.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param defval Default value for property
+     * @return Long value of or default value.
```
olenagerasimova, line 33: and this one too

src/main/java/com/artipie/misc/Property.java
```diff
@@ -0,0 +1,53 @@
...
+     * @throws ArtipieException In case of problem with parsing value of the property
+     */
+    public long asLongOrDefault(final String defval) {
```
olenagerasimova, line 36: I'd suggest to change type of the parameter from String to long, as this method is designed to obtain and return longs

olenagerasimova (changes requested):
@genryxy thanks, added some comments, take a look, please

genryxy:
@olenagerasimova thanks, corrected

olenagerasimova (approved):
@genryxy thanks

olenagerasimova:
@g4s8 please, take a look/merge

PR #964, opened 2021-11-08
feat: add cache for credentials configuration

### LINKED ISSUE #87: Unnecessary operations on each request

genryxy:
Part of #87 
Added cache for credentials configuration. At first, it reads credentials section from Artipie settings. If this content exists in cache and is not expired, credentials will be loaded from the cache. Otherwise a new instance will be created. Next I'm planning to add more tests

### DIFF
--- a/src/main/java/com/artipie/cache/CachedCreds.java
+++ b/src/main/java/com/artipie/cache/CachedCreds.java
@@ -4,33 +4,33 @@
  */
 package com.artipie.cache;
 
-import com.artipie.ArtipieException;
-import com.artipie.Settings;
-import com.artipie.UsersFromStorageYaml;
+import com.amihaiemil.eoyaml.Yaml;
+import com.amihaiemil.eoyaml.YamlInput;
+import com.amihaiemil.eoyaml.YamlMapping;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
-import com.artipie.http.slice.KeyFromPath;
-import com.artipie.management.Users;
+import com.artipie.asto.ext.PublisherAs;
+import com.artipie.asto.misc.UncheckedIOFunc;
 import com.artipie.misc.ArtipieProperties;
 import com.artipie.misc.Property;
+import com.artipie.repo.ConfigFile;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import java.util.Objects;
-import java.util.Optional;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
-import org.apache.commons.lang3.tuple.Pair;
 
 /**
  * Implementation of cache for credentials with similar configurations
  * in Artipie settings using {@link LoadingCache}.
  * @since 0.23
  */
-final class CachedCreds implements CredsConfigCache {
+public final class CachedCreds implements CredsConfigCache {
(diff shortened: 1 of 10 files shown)


src/main/java/com/artipie/cache/CachedCreds.java
```diff
@@ -0,0 +1,142 @@
...
+ * @since 0.23
+ */
+public final class CachedCreds implements CredsConfigCache {
+    /**
+     * Cache for storages settings.
+     */
+    private static LoadingCache<Metadata, CompletionStage<YamlMapping>> creds;
+
+    static {
+        CachedCreds.creds = CacheBuilder.newBuilder()
```
olenagerasimova, line 36: I'm not a fun of static section, for example, it does not allow to properly test the class: if `LoadingCache` would be normal private field and corresponding ctor, `invalidateAll` could be checked very easy... Don't you doncider changing approach?

olenagerasimova (changes requested):
@genryxy thanks, one question

src/main/java/com/artipie/cache/CachedCreds.java
genryxy: It sounds reasonable because now an instance of class is created on the top level only once (e.g. the similar way of initialization of a static field). And it is made in ctor of `SettingsCaches.All`

genryxy:
@olenagerasimova thanks, corrected and answered. I think it would be better to refactor in other places in the next PR

genryxy:
@olenagerasimova in this case some logic for loading from cache is introduced. I could just copy creating of this kind of cache to the `setUp` scope in tests but it doesn't look good. Probably it would be better to make this field package-private and get access to this field in the test scope?

olenagerasimova:
@genryxy you could use `Cache` instead of loading `LoadingCache`, thus loading logic will be introduced in `credentials` method instead of ctor and ctor will be code free as it supposed to. Cache [have](https://github.com/google/guava/wiki/CachesExplained#from-a-callable) `get` method that excepts `callable` to load and add value to the cache if it's not present.

genryxy:
@olenagerasimova thanks, made as you suggested

olenagerasimova (approved):
@genryxy thanks, now it looks much better

g4s8:
@genryxy integration tests are failing here

genryxy:
@g4s8 now it works fine. There was a problem with `CondaITCase` but after restarting it worked in a proper way

PR #967, opened 2021-11-11
refactor: use instance field for cache

### LINKED ISSUE #87: Unnecessary operations on each request

genryxy:
Closes #87
Used instance field instead of static because this instance is created only once. I didn't refactor static field in `RepositoriesFromStorage` because instances of this class are created many times

### DIFF
--- a/src/test/java/com/artipie/cache/CachedUsersTest.java
+++ b/src/test/java/com/artipie/cache/CachedUsersTest.java
@@ -5,10 +5,14 @@
 package com.artipie.cache;
 
 import com.artipie.http.auth.Authentication;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.core.IsEqual;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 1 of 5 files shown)


olenagerasimova (approved):
@genryxy thanks

genryxy:

genryxy:

src/test/java/com/artipie/cache/CachedUsersTest.java
```diff
@@ -67,14 +85,18 @@ void failsToGetUserWhenCacheContainsAnotherUser() {
...
             user, "", (usr, pass) -> Optional.of(new Authentication.User(user))
         );
         MatcherAssert.assertThat(
             target.user(user, "", (usr, pass) -> Optional.empty()).isPresent(),
             new IsEqual<>(true)
         );
+        MatcherAssert.assertThat(
+            "Size of cache differed from 1",
+            this.cache.size(),
+            new IsEqual<>(1L)
```
g4s8, line 99: It's a minor issue, but you can use [Matchers.hasSize](http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/Matchers.html#hasSize-int-) next time.

g4s8 (approved):

PR #975, opened 2021-12-12
feat: collect metrics with prometheus

### LINKED ISSUE #889: Monitoring and metrics tool

baudoliver7:
Closes #889

- We decided to collect metrics by [pull way](https://prometheus.io/docs/practices/instrumentation/#offline-processing) because of these [limitations](https://prometheus.io/docs/practices/pushing/#should-i-be-using-the-pushgateway) of `Pushgateway`.
- We store periodically metrics in a cache storage, the time that `Prometheus` pull them via `PromuSlice`. To implement `PromuSlice`, we followed [that](https://github.com/prometheus/client_java/blob/65ca8bd19382c4f35f7f8d10e2cc462faf3adf3c/simpleclient_vertx/src/main/java/io/prometheus/client/vertx/MetricsHandler.java#L73) of `MetricsHandler` for `Vert.x`.
- To use `Prometheus` as collector of metrics, we should configure this :

```yaml
meta :
  metrics :
    type : prometheus
```
- Add a `todo` to introduce unit and integration tests for this feature.


baudoliver7:
@artipie/maintainers ready to be reviewed :-)

README.md
```diff
@@ -232,6 +232,17 @@ meta:
     interval: 5 # Publishing interval in seconds, default value is 5
 ```
 
+To collect metrics via `Prometheus`, we configure metrics like this :
+
+```yaml
+metrics :
+  type : promu
```
g4s8, line 239: Let's use full name here `prometheus`

README.md
```diff
@@ -232,6 +232,17 @@ meta:
...
+  interval : 60
```
g4s8, line 240: Interval is not needed for pull strategy

g4s8 (changes requested):
@baudoliver7 please check some minor code comments and I have one design comment: when using Prometheus it's not needed to save something in storage, keeping data in memory or temporary files will be enough: saving data in storage reduces all Prometheus pull-strategy benefits because we need to store data on some central storage (most probably server in a cluster) and then pull it back from this storage to return for pull-request. It's better using push strategy, because it will be faster and more stable (fewer requests in network); you can think about the storage as a separate system service component which we call via API. So to use pull strategy we need to keep all the metrics data in memory or some temporary disk files not in storage.

README.md
baudoliver7: Yes, to be more human readable :)

README.md
baudoliver7: You are right, even if we need to save periodically metrics in memory. Ok, I will use the default interval which 5.

baudoliver7: You are right, even if we need to save periodically metrics in memory. Ok, I will use the default interval which 5.

baudoliver7:
@g4s8 I agree with you for your design comment about performance. I choose now to keep data in memory by using `InMemoryStorage` which is faster.

baudoliver7:
@g4s8 I suggest another point of view. We do not need to save metrics somewhere. We can simply use `Metrics` instance in `PromuSlice` by output there all available metrics stored in. WDYT ?

g4s8:
> @g4s8 I suggest another point of view. We do not need to save temporary metrics somewhere. We can simply use `Metrics` instance in `PromuSlice` by output directly there all available metrics stored in. This will ensure in this way that `Prometheus` retrieves the latest values of the metrics. WDYT ?

@baudoliver7 I didn't get what did you mean here, can you clarify please:
> Metrics instance in PromuSlice by output directly

baudoliver7:
@g4s8 I mean do this :

```java
/**
 * New slice with metrics.
 * @param metrics Metrics
 */
public PromuSlice(final Metrics metrics) {
    this.metrics = metrics;
}

@Override
public Response response(final String line, final Iterable<Map.Entry<String, String>> headers,
    final Publisher<ByteBuffer> body) {
    // We output metrics directly here instead of saving them temporarily somewhere else.
    final MetricsOutput output = new StorageMetricsOutput(
        new InMemoryStorage()
    );
    metrics.publish(output);
    // send them here
    ...
    ...
}
```

baudoliver7:
@g4s8 fine ! Let's do it :-)

baudoliver7:
@g4s8 please, take a look at my latest changes :)

olenagerasimova (dismissed):
@baudoliver7 looks good to me

src/main/java/com/artipie/metrics/publish/PrometheusOutput.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param data Metrics to write
+     */
+    private void write(final Map<String, Long> data) {
+        for (final Map.Entry<String, Long> metric : data.entrySet()) {
+            // @checkstyle MethodBodyCommentsCheck (1 line)
+            // @see https://github.com/prometheus/client_java#counter
+            Counter.build()
+                .name(metric.getKey())
+                .register(this.registry)
+                .inc(metric.getValue());
```
g4s8, line 77: It's not correct: Prometheus has different data types for counters and gauges: https://prometheus.io/docs/concepts/metric_types/ - but this code adds gauges to counters as increments, so it'll add gauges to previous gauge values instead of overwrite it.

g4s8 (changes requested):
@baudoliver7 please one comment about implementation issue

src/main/java/com/artipie/metrics/publish/PrometheusOutput.java
baudoliver7: well seen !

baudoliver7:
@g4s8 I think it's ok now.

g4s8 (approved):
@baudoliver7 great, thanks!

Issue #978, opened 2021-12-15
PromuSlice.java:30-34: Test that {@link PromuSlice}'s...


0pdd:
The puzzle `889-be863331` from #889 has to be resolved: 

https://github.com/artipie/artipie/blob/90c304bb9075bd02e5ca97ecfcb078b90ee4ba79/src/main/java/com/artipie/http/PromuSlice.java#L30-L34

The puzzle was created by @baudoliver7 on 15-Dec-21. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

baudoliver7:
@g4s8 could you assign me this please ?

baudoliver7:
@g4s8 could you assign me this please ?

0pdd:
@0pdd the puzzle [#980](https://github.com/artipie/artipie/issues/980) is still not solved.

PR #984, opened 2022-01-13
refactor: replace params method in PromuSlice

baudoliver7:
A new method has been introduced in `RqParams` (artipie/http) that allow us to obtain all values of a
query parameter. We refactored `PromuSlice` with this new method.

Ref: #978

### DIFF
--- a/src/main/java/com/artipie/http/PromuSlice.java
+++ b/src/main/java/com/artipie/http/PromuSlice.java
@@ -20,7 +20,6 @@ import java.io.StringWriter;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import org.reactivestreams.Publisher;
 
--- a/src/main/java/com/artipie/http/PromuSlice.java
+++ b/src/main/java/com/artipie/http/PromuSlice.java
@@ -9,22 +9,19 @@ import com.artipie.asto.Content;
 import com.artipie.http.headers.Accept;
 import com.artipie.http.headers.ContentType;
 import com.artipie.http.rq.RequestLineFrom;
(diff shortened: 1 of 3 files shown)


baudoliver7:
@artipie/maintainers ready to be reviewed :)

genryxy (approved):
@baudoliver7 thanks, looks good to me
@g4s8 please check

baudoliver7:
@g4s8 waiting to be merged

src/main/java/com/artipie/http/PromuSlice.java
```diff
@@ -54,8 +51,12 @@ public Response response(final String line, final Iterable<Map.Entry<String, Str
...
         final String ctype = new Accept(headers).values().get(0);
         try (StringWriter writer = new StringWriter()) {
-            final Set<String> filters = PromuSlice.params(line, "name");
-            final MetricsOutput output = new PrometheusOutput(writer, ctype, filters);
+            final List<? extends String> filters = new RqParams(
+                new RequestLineFrom(line).uri()
+            ).values("name");
+            final MetricsOutput output = new PrometheusOutput(
+                writer, ctype, new HashSet<>(filters)
+            );
```
g4s8, line 59: It seems we use `filters` only once, I'd suggest inline this variable

g4s8 (changes requested):
@baudoliver7 thanks, please check my comment

src/main/java/com/artipie/http/PromuSlice.java
baudoliver7: fixed

baudoliver7:
@g4s8 I took advantage to bump `pdd-action` from `v1.0.0` to `master`. All things look good now :)

g4s8 (approved):
@baudoliver7 thanks

Issue #273, opened 2020-07-07
ResponseMetricsSlice.java:45-48: Report exceptions as...


0pdd:
The puzzle `231-bfc96007` from #231 has to be resolved: 

https://github.com/artipie/artipie/blob/36a1c5a395adebe1d657e1f55718d47e66666eb4/src/main/java/com/artipie/ResponseMetricsSlice.java#L45-L48

The puzzle was created by @olegmoz on 07-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8), @olenagerasimova[/z](https://www.0crat.com/u/olenagerasimova) please, pay attention to this issue

PR #987, opened 2022-01-20
fix: report exceptions as error in ResponseMetricsSlice

baudoliver7:
We catch exceptions and send them to metrics.

Fix: #273

### DIFF
--- a/src/main/java/com/artipie/ResponseMetricsSlice.java
+++ b/src/main/java/com/artipie/ResponseMetricsSlice.java
@@ -23,10 +23,6 @@ import org.reactivestreams.Publisher;
  * Returned response reported to metrics as success or error if it's status code is 4xx or 5xx.
  *
  * @since 0.9
- * @todo #231:30min Report exceptions as errors in `ResponseMetricsSlice`.
- *  In case of exceptions in origin slice handling or sending to `Connection` nothing is reported
- *  to `Metrics`. It needs to be fixed, so if origin slice fails or we failed to send response
- *  then error is sent to `Metrics`.
  */
 public final class ResponseMetricsSlice implements Slice {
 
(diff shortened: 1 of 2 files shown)


baudoliver7:
@artipie/maintainers ready to be reviewed :)

baudoliver7:
@g4s8 I think you forgot to review this one after assigning :)

src/main/java/com/artipie/ResponseMetricsSlice.java
```diff
@@ -51,16 +48,25 @@ public ResponseMetricsSlice(final Slice origin, final Metrics metrics) {
...
+        return connection -> {
+            try {
+                return this.origin.response(rqline, rqheaders, rqbody).send(
+                    (rsstatus, rsheaders, rsbody) -> {
+                        this.report(rqline, rqheaders, rsstatus);
+                        return connection.accept(rsstatus, rsheaders, rsbody);
+                    }
+                );
+            // @checkstyle IllegalCatchCheck (1 line)
+            } catch (final Exception exc) {
```
g4s8, line 65: We need to catch `ArtipieHttpException` first, since it can include some error status and message, an generic exception only after that:
```java
} catch (ArtipieHttpException ex) {
  // report with ex.status
} catch (Exception ex) {
  // report INTERNAL_ERROR
}
```

g4s8 (changes requested):
@baudoliver7 thanks, please check my comment

src/main/java/com/artipie/ResponseMetricsSlice.java
baudoliver7: Ok

g4s8 (approved):
@baudoliver7 thanks

Issue #977, opened 2021-12-15
integration-tests.yml:32-37: Fix integration tests and...


0pdd:
The puzzle `855-2061b0aa` from #855 has to be resolved: 

https://github.com/artipie/artipie/blob/90c304bb9075bd02e5ca97ecfcb078b90ee4ba79/.github/workflows/integration-tests.yml#L32-L37

The puzzle was created by Evgeny Chugunnyy (John) on 10-Dec-21. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@0pdd the puzzle [#996](https://github.com/artipie/artipie/issues/996) is still not solved.

0pdd:
@0pdd the puzzle [#1013](https://github.com/artipie/artipie/issues/1013) is still not solved; solved: [#996](https://github.com/artipie/artipie/issues/996).

0pdd:
@0pdd all 2 puzzles are solved here: [#1013](https://github.com/artipie/artipie/issues/1013), [#996](https://github.com/artipie/artipie/issues/996).

PR #989, opened 2022-01-21
test: fix VertxMainITCase

baudoliver7:
We migrate `VertxMainITCase` to testcontainers and we fix test `startsWhenNotValidRepoConfigsArePresent` that was failing.
We add a puzzle to fix other test case.

Fix: #977

### DIFF
--- a/src/main/java/com/artipie/RepoConfig.java
+++ b/src/main/java/com/artipie/RepoConfig.java
@@ -34,7 +34,7 @@ import org.reactivestreams.Publisher;
  * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  * @checkstyle ParameterNumberCheck (500 lines)
  */
-@SuppressWarnings("PMD.TooManyMethods")
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.AvoidDuplicateLiterals"})
 public final class RepoConfig {
 
     /**
@@ -199,8 +199,10 @@ public final class RepoConfig {
      * @return Async YAML mapping
      */
     public YamlMapping repoConfig() {
-        return Optional.ofNullable(this.yaml.yamlMapping("repo")).orElseThrow(
-            () -> new IllegalStateException("Invalid repo configuration")
+        return Optional.ofNullable(this.yaml.yamlMapping("repo")).orElse(
+            Yaml.createYamlMappingBuilder()
+                .add("type", "file")
+                .add("storage", "default").build()
         );
     }
 
--- a/src/main/java/com/artipie/RepoConfig.java
+++ b/src/main/java/com/artipie/RepoConfig.java
@@ -199,10 +199,8 @@ public final class RepoConfig {
      * @return Async YAML mapping
      */
     public YamlMapping repoConfig() {
-        return Optional.ofNullable(this.yaml.yamlMapping("repo")).orElse(
-            Yaml.createYamlMappingBuilder()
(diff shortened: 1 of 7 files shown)


src/main/java/com/artipie/RepoConfig.java
```diff
@@ -199,8 +199,10 @@ public Storage storage() {
      * @return Async YAML mapping
      */
     public YamlMapping repoConfig() {
-        return Optional.ofNullable(this.yaml.yamlMapping("repo")).orElseThrow(
-            () -> new IllegalStateException("Invalid repo configuration")
+        return Optional.ofNullable(this.yaml.yamlMapping("repo")).orElse(
+            Yaml.createYamlMappingBuilder()
```
g4s8, line 203: I don't think it's a good solution to implicitly create configuration for repo in `src/main` - it may cause unpredictable behavior for users. I'd be better to set up proper configuration in for test.

src/test/java/com/artipie/VertxMainITCase.java
```diff
@@ -27,7 +26,6 @@
  * @since 0.1
  */
 @SuppressWarnings("PMD.AvoidDuplicateLiterals")
-@Disabled("FIXME: this test is failing on JVM-17, checks the logs")
```
g4s8, line 30: We also need to move this test to testcontainers (like other working tests here) as described in todo.

g4s8 (changes requested):
@baudoliver7 please check my comments

src/main/java/com/artipie/RepoConfig.java
baudoliver7: Ok

src/test/java/com/artipie/VertxMainITCase.java
baudoliver7: Ok

baudoliver7: Good now :)

src/main/java/com/artipie/RepoConfig.java
baudoliver7: @g4s8 I think it's normal to respond `500` when we haven't a valid repo config. So `startsWhenNotValidRepoConfigsArePresent` should rather expect `500` instead of `200`. WDYT ?

baudoliver7:
@g4s8 please, see the latest changes

g4s8 (approved):
@baudoliver7 thanks

baudoliver7:
@g4s8 Auto-merging actually failing because of a trouble with `RpmITCase` while checking `integration-tests` :
```console
Error:    RpmITCase.setUp:45 Yum install curl failed
Expected: status <0>stdout an instance of java.lang.String
     but: failed with status:
<1>
STDOUT: CentOS Linux 8 - AppStream                      150  B/s |  38  B     00:00    

STDERR: Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```

Could I disable this test and report this issue in a separate ticket ?

baudoliver7:
@g4s8 Bug about `RpmITCase` reported in issue #993.
I disabled the test. You can merge it now :)

g4s8 (approved):

baudoliver7:
@g4s8 I noticed that puzzle has not been generated because of a `@todo` in `CHANGELOG.md`.
I suggested this PR (#994)  to fix it.

PR #992, opened 2022-01-30
feat: alternative authentication

### LINKED ISSUE #337: Settings configuration for GitHub auth.
The puzzle `285-66616a94` from #285 has to be resolved: 

https://github.com/artipie/artipie/blob/36d95233470090223c491a89f945f48ad1f55b8c/src/main/java/com/artipie/YamlSettings.java#L50-L55

The puzzle was created by @g4s8 on 15-Jul-20. 

dgarus:
Closes #337 

- added an opportunity to set Github auth in YAML settings
- added an opportunity to set a few authentications in YAML settings
```
 credentials:
    -
      type: file
      path: _credentials.yaml
    -
      type: env
```
If there is only one type of credentials, then it can be set up as before
```
 credentials:
      type: file
      path: _credentials.yaml
```
- added tests to check Github auth and alternative authentication

### DIFF
--- a/src/main/java/com/artipie/UsersFromGithub.java
+++ b/src/main/java/com/artipie/UsersFromGithub.java
@@ -15,7 +15,7 @@ import java.util.concurrent.CompletionStage;
  * Credentials from GitHub.
  * @since 0.12.2
  */
-public class UsersFromGithub implements Users {
+public final class UsersFromGithub implements Users {
     @Override
     public CompletionStage<List<User>> list() {
         return CompletableFuture.failedFuture(
--- a/src/main/java/com/artipie/UsersFromGithub.java
+++ b/src/main/java/com/artipie/UsersFromGithub.java
@@ -1,0 +1,48 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/artipie/LICENSE.txt
+ */
+package com.artipie;
+
+import com.artipie.auth.GithubAuth;
+import com.artipie.http.auth.Authentication;
+import com.artipie.management.Users;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * Credentials from GitHub.
+ * @since 0.12.2
+ */
+public class UsersFromGithub implements Users {
+    @Override
+    public CompletionStage<List<User>> list() {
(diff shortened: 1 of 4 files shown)


dgarus:
@g4s8 hi!
Could you please review the PR?

dgarus:
@artipie/maintainers Hello!

olenagerasimova (commented):
@dgarus thanks, code looks good to me. Though, I'd suggest to organize authentications as a yaml list instead of nested maps:
```yaml
credentials:
  - type: env
  - type: github
  - type: file
    path: _credentials.yaml
```
@g4s8 please, take a look. What do you think?

dgarus:
@olenagerasimova @g4s8 
When we use a list, all elements are the same, 
but there is a difference between credentials and authentication (that is only part of credentials), 
so I suggest separating their setup.

g4s8:
I agree with @olenagerasimova - first: nested configuration file looks overcomplicated; second: we may have multiple configurations with the same type, e.g. `[{type: file, path: common.txt}, {type: file, path: private.txt}]`.

Also, @dgarus I didn't get what the problem is
> When we use a list, all elements are the same,
but there is a difference between credentials and authentication (that is only part of credentials)

YAML list items are dynamic, no need to use same structure for each item.

dgarus:
@g4s8 @olenagerasimova 
I see one problem:
What should retun the `com.artipie.YamlSettings#credentials()` when the field `credentials:` is a list of credentials?
Should it be the first element of the list?

g4s8:
> What should retun the `com.artipie.YamlSettings#credentials()` when the field `credentials:` is a list of credentials? Should it be the first element of the list?

@dgarus it could be an ordered list or some object which encapsulates this list and can authenticate request using this list. In you example you are using almost the same approach:
```
credentials:
    type: env
    alternative_auth:
      - type: github
      - type: file
        path: _credentials.yaml
```
So the implementation could be very similar. But I don't understand why one credential's source should be a part of another (semantically).

dgarus:
@olenagerasimova @g4s8 
Hello!
I fixed the YAML configuration of credentials.
Please, have a look.

g4s8 (approved):
@dgarus thanks!

baudoliver7:
@g4s8 Please, could you set `pdd-lint` to `required` ? If you check [details](https://github.com/artipie/artipie/runs/5067182107?check_suite_focus=true), you will see that `pdd` fail because of a short description.

PR #997, opened 2022-02-08
fix: corrected RpmITCase to use fedora instead of sentos 8

### LINKED ISSUE #993: RpmITCase fails on set up
We notice that `RpmITCase` fails with this error message :
```console
Error:    RpmITCase.setUp:45 Yum install curl failed
Expected: status <0>stdout an instance of java.lang.String
     but: failed with status:
<1>
STDOUT: CentOS Linux 8 - AppStream                      150  B/s |  38  B     00:00    

STDERR: Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```

We have to fix it and enable the test.

olenagerasimova:
Closes #993 
Sentos 8 went EOL in December 2021, so I've corrected the test to use fedora 35 and dnf.


g4s8 (approved):
@olenagerasimova :+1:

PR #998, opened 2022-02-08
decrease min-words from 20 to 16

portlek:
this pr fixes pdd issue about this [todo](https://github.com/artipie/artipie/blob/master/src/main/java/com/artipie/YamlSettings.java#L29) which has 16 only words although the minimum words for todo is 20.
the another solution would be increasing words of the todo.


g4s8 (approved):
@portlek thanks

PR #1000, opened 2022-02-09
fix: Updated python and conda adapters

olenagerasimova:
Updated python and conda adapters which now uses new multipart sdk from http. Also, updated asto and applied necessary changes.

Upd: updates python IT to upload only one package. Not sure why, but uploading several files gets stuck. From experience with pypi-adapter IT it should be caused by the environment, so I've changed the test to upload one file and added test with multiple files into the adapter https://github.com/artipie/pypi-adapter/pull/168


olenagerasimova:
@g4s8 http fix with executor pull helped, please, merge

g4s8 (approved):
@olenagerasimova thanks

PR #1001, opened 2022-02-09
test: DockerProxyIT migrated to testcontainers

### LINKED ISSUE #996: integration-tests.yml:32-36: Fix integration tests and...
The puzzle `977-02d95599` from #977 has to be resolved: 

https://github.com/artipie/artipie/blob/5dd565f67b264e3aaffe40b22ecc7d6d48609172/.github/workflows/integration-tests.yml#L32-L36

The puzzle was created by @baudoliver7 on 03-Feb-22. 

dgarus:
Closes #996 

- Migrated DockerProxyIT to test containers
- added a puzzle to refactor tests that require executing several steps like DockerProxyIT or DockerLocalITCase
- add a puzzle to fix DockerOnPortIT test

### DIFF
--- a/src/test/java/com/artipie/docker/DockerProxyIT.java
+++ b/src/test/java/com/artipie/docker/DockerProxyIT.java
@@ -4,22 +4,18 @@
  */
 package com.artipie.docker;
 
-import com.amihaiemil.eoyaml.Yaml;
-import com.artipie.ArtipieServer;
-import com.artipie.RepoConfigYaml;
-import com.artipie.docker.junit.DockerClient;
-import com.artipie.docker.junit.DockerClientSupport;
 import com.artipie.docker.proxy.ProxyDocker;
-import java.nio.file.Path;
-import org.hamcrest.MatcherAssert;
-import org.hamcrest.core.StringContains;
-import org.junit.jupiter.api.AfterEach;
+import com.artipie.test.ContainerResultMatcher;
+import com.artipie.test.TestDeployment;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.List;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledOnOs;
 import org.junit.jupiter.api.condition.OS;
-import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import wtf.g4s8.tuples.Pair;
 
 /**
  * Integration test for {@link ProxyDocker}.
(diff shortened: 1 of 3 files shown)


dgarus:
@artipie/maintainers 

src/test/java/com/artipie/docker/DockerProxyIT.java
```diff
@@ -41,75 +37,70 @@
  *  It seems that body is being read by some other entity in Artipie,
  *  so it requires investigation.
  *  Similar `CachingProxyITCase` tests works well in docker-adapter module.
+ *  @todo #996:30min Refactor set up of a test's steps.
+ *   Consider creating a class to set up steps of test using less verbose way.
```
olenagerasimova, line 41: I'm not sure what exactly requires refactoring in set up - there is just one line.

src/test/java/com/artipie/docker/DockerProxyIT.java
```diff
@@ -41,75 +37,70 @@
...
+                    "artipie:8080"
+                )
+            ),
+            Pair.of(
+                "Failed to pull image",
+                List.of(
+                    "docker", "pull", img
+                )
+            )
+        ).forEach(
```
olenagerasimova, line 90: There are several consernes:
- it's not easy to read and understand what is going on here with such approach. The best tests are plain, simple and obvious, with no complications (as each complication is a potential place for error, and each error in test can lead to false positive/negative affect). Thus, I'd suggest to get rid of the list and just call `this.deployment.assertExec()` twice with different arguments. I know that all docker tests are written this way, but do not think this is a good approach.
- exec code 0 does not actually mean the success here - are we sure the image was downloaded from artipie? I understand, that it was demanded in pull command, but we should consider the possibility that command was written incorrectly or for some reason it did not work and image was loaded from global registry. Thus, if possible, I'd suggest to add some expected strings into matcher (check conda test as an example). Also, some package managers return 0 exit code even package was not found

olenagerasimova (changes requested):
@dgarus thanks, check my comments please

src/test/java/com/artipie/docker/DockerProxyIT.java
dgarus: I added a code example of how it can look. 
So if you think this is excess, I'll remove the puzzle.

dgarus:
@olenagerasimova, thank you very much for the review!
I fixed your comments. 
Have please a look.

olenagerasimova (dismissed):
@dgarus thanks
@g4s8 please, merge

g4s8 (approved):
@dgarus thanks

g4s8:
@dgarus BTW puzzle linter failed and blocks this PR, see logs: https://github.com/artipie/artipie/runs/5140189302?check_suite_focus=true

g4s8 (approved):

PR #1002, opened 2022-02-09
fix: migrate rpm smoke test for fedora

olenagerasimova:
Part of #993 
Fixed rpm smoke test by migrating it to fedora (sentos 8 went EOL in December 2021)


g4s8 (approved):
@olenagerasimova thanks

PR #1004, opened 2022-02-11
test: integration test for Docker proxy push feature

### LINKED ISSUE #511: DockerProxyIT.java:51-53: Add integration test for Docker...
The puzzle `499-4f132183` from #499 has to be resolved: 

https://github.com/artipie/artipie/blob/ab444fd0b00d0dbe4240648afb1f6fdd8ad95fc2/src/test/java/com/artipie/docker/DockerProxyIT.java#L51-L53

The puzzle was created by Oleg Mozzhechkov on 25-Aug-20. 

dgarus:
Closes #511


dgarus:
@artipie/maintainers Hello!

src/test/java/com/artipie/docker/DockerProxyIT.java
```diff
@@ -100,4 +97,60 @@ void shouldPullRemote() throws Exception {
...
+        this.deployment.assertExec(
+            "Failed to push image to Artipie",
+            new ContainerResultMatcher(
+                new IsEqual<>(ContainerResultMatcher.SUCCESS),
+                new StringContains(
+                    "The push refers to repository [artipie:8080/my-docker/alpine]"
+                )
+            ),
+            "docker", "push", image
+        );
```
g4s8, line 136: Why does it work? You are pushing image to proxy registry with `registry-1.docker.io` and `mcr.microsoft.com` remotes, but the proxy doesn't have push permissions in these remotes, isn't it?

g4s8 (changes requested):
@dgarus one questions

src/test/java/com/artipie/docker/DockerProxyIT.java
dgarus: Sure it doesn't have; it pushes an image to the local storage as declared in the [doc](https://github.com/artipie/artipie/tree/master/examples/docker#docker-proxy-repo).
The task description is 
`Docker proxy supports pushing to local storage if such storage is specified. 
 It should be verified that an image can be pushed to a proxy repository and pulled later`.
It looks like we should test `pushing to local storage,` so I think all is ok.

g4s8 (approved):
@dgarus thanks

Issue #995, opened 2022-02-04
Smoke-tests checks always fail


baudoliver7:
Two tests fail (`conda` and `rpm`) according to this output :
```console
all tests finished:
+ grep FAILED /home/runner/work/artipie/artipie/examples/results.txt
test binary - PASSED
test debian - PASSED
test docker - PASSED
test go - PASSED
test helm - PASSED
test maven - PASSED
test npm - PASSED
test nuget - PASSED
test php - PASSED
test rpm - FAILED
test conda - FAILED
```

### RPM test

Here is the output when `RPM` test fails :
```console
+ dnf -y repository-packages example install
Example Repository                              7.0 kB/s | 253  B     00:00    
Fedora 35 - x86_64                               19 MB/s |  79 MB     00:04    
Fedora 35 openh264 (From Cisco) - x86_64        3.2 kB/s | 2.5 kB     00:00    
Fedora Modular 35 - x86_64                      2.5 MB/s | 3.3 MB     00:01    
Fedora 35 - x86_64 - Updates                     12 MB/s |  27 MB     00:02    
Fedora Modular 35 - x86_64 - Updates            935 kB/s | 2.9 MB     00:03    
No package available.
Error: No packages marked for install.
+ [[ 1 == \0 ]]
+ tee -a /home/runner/work/artipie/artipie/examples/results.txt
+ echo 'test rpm - FAILED'
```

### CONDA test

Here is the output when `CONDA` test fails :
```console
java.util.concurrent.CompletionException: com.artipie.asto.ArtipieIOException: java.io.IOException: Entry path is out of storage: my-conda/./upload/linux-64/snappy-1.1.3-0.tar.bz2
java.util.concurrent.CompletionException: com.artipie.asto.ArtipieIOException: java.io.IOException: Entry path is out of storage: my-conda/./upload/linux-64/snappy-1.1.3-0.tar.bz2
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
	at java.base/java.util.concurrent.CompletableFuture.uniApplyNow(CompletableFuture.java:674)
	at java.base/java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:662)
	at java.base/java.util.concurrent.CompletableFuture.thenApplyAsync(CompletableFuture.java:2173)
	at com.artipie.asto.fs.FileStorage.exists(FileStorage.java:76)
[...]

PR #1015, opened 2022-02-18
fix(smoke-rpm): add repository `example` to `dnf`

baudoliver7:
Repository `example` was defined but not added. We simply added like it is described [here](https://www.tecmint.com/configure-software-repositories-in-fedora/).

Ref: #995


olenagerasimova:
@g4s8 please, merge

g4s8 (approved):

PR #1017, opened 2022-02-25
fix: removed unused classes

olenagerasimova:
As we fully migrated to docker containers in IT I've removed unused classes previously required for ITs.


g4s8 (approved):
@olenagerasimova thanks

PR #1029, opened 2022-06-09
deps: bump eo-yaml to 6.0.1

olenagerasimova:
Updated eo-yaml to 6.0.1. Previous version worked incorrectly (simply ignored second and other sequence items) with sequence of mappings, formatted like this
```yaml
credentials:
    - type: file
      path: _credentials.yml
    - type: github
```
New version is a bit better, it requires new lines only when value contain `-` sign:
```yaml
remotes:
    -
      url: http://artipie:8080/my-maven
      username: alice
      password: 123
```

Also, this version fails, when yaml sequence is tried to obtain by special not-quoted symbol `*`, that's why I added quoted constant in `YamlPermissions`. Also, fixed data for `DockerProxyTest`


olenagerasimova:
too many changes, will not merge

PR #1030, opened 2022-06-15
feat: add docker-compose.yaml and correct README.md

### LINKED ISSUE #1028: Change Artipie release workflow to include front service
As dashboard was moved to separate repository and runs separately now, we have to change Artipie release script and provide Artipie as service with dashboard included. There seems to be two approaches to achieve that:
- provide some docker-compose with Artipie and `front` service as docker images
- include `front` jar file into Artipie docker image

After preparing new release, if necessary, we have to change "Quickstart" section of the readme file accordingly and check other places: wiki, [CONTRIBUTING.md](https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md) and other docs.

olenagerasimova:
Closes #1028 
Added docker-compose script to run both Artipie and Artipie front, corrected readme. 

Note, that I'm not a devops and have no idea about best practicies for such type of configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,10 +50,12 @@ docker-compose up
 ```
 
 It'll start a new Docker container with latest Artipie and Artipie dashboard service image. 
+Containers should share same config directory, default local mount location is `/usr/local/artipie`,
+you may need to correct it if docker client is running not on linux operating system.
 A new image generate default configuration if not found at `/etc/artipie/artipie.yml`, prints initial
 credentials to console and prints a link to the dashboard. If started on localhost with command
 above, the dashboard URI is `http://localhost:8080/dashboard/artipie` and default username and password 
-are `artipie/artipie`. Artipie server side (repositories) are served on `8081` port and are 
+are `artipie/artipie`. Artipie server side (repositories) is served on `8081` port and is 
 available on URI `http://localhost:8081/artipie/{reponame}`, where `{reponame}` is the name of the
 repository.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -44,31 +44,35 @@ The following set of features makes Artipie unique among all others:
 # Quickstart
```
dgarus, line 44: Currently, we cannot host the data in Google Cloud and HuaweiCloud OBS...

dgarus (changes requested):
@olenagerasimova, pay attention to my comment, please.

dgarus (dismissed):
Thanks!
LGTM

olenagerasimova:
@dgarus thanks, I've corrected part about storages. Have you tried to run docker-compose? Steps to check everything is working can be the following:
- run docker-compose
- go to dashboard: 
    * login with test user `artipie/artipie`
    * create new binary repository
- put some file into created repository (use `curl PUT` or smth similar)
- get the file from repository (use `curl GET` or smth similar)

olenagerasimova:
@dgarus I've corrected docker-compose script, storages configuration and readme. 
As for [named volumes](https://docs.docker.com/storage/volumes/#start-a-service-with-volumes) 

> When you start a service and define a volume, each service container uses its own local volume. None of the containers can share this data if you use the local volume driver

Volumes also has some volumes drivers, but most of them are for AWS or Azure or some other cloud storage support, not for simple data sharing. 

That's why I simply added two volumes with some (linux consistent) local directory.

PR #1033, opened 2022-06-20
fix: casesensitive username in auth from github

### LINKED ISSUE #1032: Github auth & permissions case sensitivity mismatch
Github auth currently may return login string with different case, which makes permissions setup more difficult, since they must match internal GitHub case. Possible solution is to make it case-insensitive, but let's discuss it first.

Example for `github.com/ChGen` login:
```
[INFO] 2022-06-16 13:35:39,161 ForkJoinPool.commonPool-worker-1 com.artipie.http.auth.Authentication$Joined - Successfully authenticated 'github.com/ChGen' user via Joined([AuthFromYaml(),Cached(GithubAuth())])
[INFO] 2022-06-16 13:35:39,167 ForkJoinPool.commonPool-worker-1 com.artipie.YamlPermissions - Operation 'deploy' denied for 'chgen'
[INFO] 2022-06-16 13:35:39,168 ForkJoinPool.commonPool-worker-1 com.artipie.YamlPermissions - Operation 'upload' allowed for 'chgen'
[INFO] 2022-06-16 13:35:39,182 ForkJoinPool.commonPool-worker-1 com.artipie.asto.fs.FileStorage - Save 'ChGen/bintest/artipie.yml': Optional[153]
[INFO] 2022-06-16 13:35:39,183 ForkJoinPool.commonPool-worker-1 com.artipie.http.MainSlice - << CREATED
```

So, it receives 'chgen' internally, and repository permissions currently must match GitHub:

`/var/artipie/repo/ChGen/bintest.yaml`
```
repo:
  type: file
  storage: default
  permissions:
  .........
    chgen:        #case sensitive!!
      - "upload"
      - "download"
```

olenagerasimova:
Closes #1032

Corrected `GithubAuth` to use case-sensitive username as it was passed by user in auth header. 
Usernames in Artipie (in `_credeentials.yaml` and use from env) are generally case-sensitive, so here we should follow the same idea. 
Github usernames are not case-sensitive, thus we compare usernames ignoring capitals and inside Artipie use case-sensitive usernames as they were passed by users. In permissions (repositories and API) usernames should case-sensitively correspond usernames, passed in auth header.

### DIFF
--- a/src/main/java/com/artipie/auth/GithubAuth.java
+++ b/src/main/java/com/artipie/auth/GithubAuth.java
@@ -70,7 +70,7 @@ public final class GithubAuth implements Authentication {
                 if (
                     Objects.equals(login, matcher.group(1).toLowerCase(Locale.US))
                 ) {
-                    result = Optional.of(new Authentication.User(login));
+                    result = Optional.of(new Authentication.User(matcher.group(1)));
                 }
             } catch (final AssertionError error) {
                 if (error.getMessage() == null
--- a/src/test/java/com/artipie/auth/GithubAuthTest.java
+++ b/src/test/java/com/artipie/auth/GithubAuthTest.java
@@ -33,7 +33,7 @@ final class GithubAuthTest {
                     return "";
                 }
             ).user("github.com/UsEr", secret).orElseThrow(),
-            new IsEqual<>(new Authentication.User("user"))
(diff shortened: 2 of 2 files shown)


olenagerasimova:
@dgarus please, take a look, this is a small but important fix

dgarus (approved):
@olenagerasimova, thanks!
LGTM

PR #1038, opened 2022-06-24
dep: updated asto and rpm

olenagerasimova:


olenagerasimova:
@dgarus please, take a look, I've only updated deps here.

dgarus (approved):
Thanks!

Issue #1037, opened 2022-06-24
Fix smoke test CI script 


olenagerasimova:
There are several problems with smoke test scripts and [examples](https://github.com/artipie/artipie/tree/master/examples) in Artipie:
1. Smoke tests CI script [fails](https://github.com/artipie/artipie/runs/6965556599?check_suite_focus=true) in spite of the fact that all tests are successful
2. In [examples](https://github.com/artipie/artipie/tree/master/examples#readme), docker and pypi are red, but actually the tests passed

ChGen:
Also fixed issues in pdd-linter:
https://github.com/g4s8/pdd-action/pull/7
https://github.com/g4s8/pdd-action/pull/8

PR #1040, opened 2022-06-29
Improving smoking tests stability

ChGen:
Improving smoking tests stability.
It quite stable now for both local/partial and full/GitHub runs.
See also: https://github.com/artipie/artipie/issues/1037

### DIFF
--- a/examples/run.sh
+++ b/examples/run.sh
@@ -139,7 +139,7 @@ function rm_volume {
 function run_test {
   local name=$1
   log_debug "running smoke test $name"
-  cd "./${name}"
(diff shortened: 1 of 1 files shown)


ChGen:
@olenagerasimova , please check. You could restart this tests in GitHub few times to see how it performs.
I still do some experimends locally.

olenagerasimova (approved):
@ChGen thanks, I've run smoke test action three times, no problems

dgarus (approved):
@ChGen thanks!
Big deal!

Issue #1041, opened 2022-07-14
Repository port


olenagerasimova:
There are two problems when repository port is specified. Let's say artipie main port is 8085 and repository `den` port is 8088, then

1) repository is available on artipie main port, that is on http://localhost:8085/den. This is not correct, NOT FOUND error should be returned in thos case

2) when repository port is specified, repository is not accessible on http://localhost:8088/den, but it's accessible without repo name by http://localhost:8088/. That is not correct either, repository should be accessible on http://localhost:8088/den

0pdd:
@olenagerasimova 10 puzzles [#1044](https://github.com/artipie/artipie/issues/1044), [#1045](https://github.com/artipie/artipie/issues/1045), [#1046](https://github.com/artipie/artipie/issues/1046), [#1047](https://github.com/artipie/artipie/issues/1047), [#1048](https://github.com/artipie/artipie/issues/1048), [#1049](https://github.com/artipie/artipie/issues/1049), [#1051](https://github.com/artipie/artipie/issues/1051), [#1052](https://github.com/artipie/artipie/issues/1052), [#1054](https://github.com/artipie/artipie/issues/1054), [#1055](https://github.com/artipie/artipie/issues/1055) are still not solved; solved: [#1043](https://github.com/artipie/artipie/issues/1043), [#1050](https://github.com/artipie/artipie/issues/1050).

0pdd:
@olenagerasimova 8 puzzles [#1045](https://github.com/artipie/artipie/issues/1045), [#1046](https://github.com/artipie/artipie/issues/1046), [#1047](https://github.com/artipie/artipie/issues/1047), [#1048](https://github.com/artipie/artipie/issues/1048), [#1049](https://github.com/artipie/artipie/issues/1049), [#1051](https://github.com/artipie/artipie/issues/1051), [#1054](https://github.com/artipie/artipie/issues/1054), [#1055](https://github.com/artipie/artipie/issues/1055) are still not solved; solved: [#1043](https://github.com/artipie/artipie/issues/1043), [#1044](https://github.com/artipie/artipie/issues/1044), [#1050](https://github.com/artipie/artipie/issues/1050), [#1052](https://github.com/artipie/artipie/issues/1052).

0pdd:
[...]

PR #1042, opened 2022-07-14
fix: repository port usage

olenagerasimova:
Part of #1041 

Fixed cases with repository port:

1) path trimming is necessary when repo port is set (for now, I've only fixed it for files repository and added corresponding tests)
2) repository with port should not be accessible by default port

### DIFF
--- a/src/main/java/com/artipie/SliceFromConfig.java
+++ b/src/main/java/com/artipie/SliceFromConfig.java
@@ -116,16 +116,12 @@ public final class SliceFromConfig extends Slice.Wrap {
         );
         switch (cfg.type()) {
             case "file":
-                slice = trimIfNotStandalone(
-                    settings, standalone,
-                    new FilesSlice(cfg.storage(), permissions, auth)
+                slice = new TrimPathSlice(
+                    new FilesSlice(cfg.storage(), permissions, auth), settings.layout().pattern()
                 );
                 break;
             case "file-proxy":
-                slice = trimIfNotStandalone(
-                    settings, standalone,
-                    new FileProxy(http, cfg)
-                );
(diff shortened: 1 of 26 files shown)


src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -116,16 +116,12 @@ static Slice build(
         );
         switch (cfg.type()) {
             case "file":
-                slice = trimIfNotStandalone(
-                    settings, standalone,
-                    new FilesSlice(cfg.storage(), permissions, auth)
+                slice = new TrimPathSlice(
```
dgarus, line 119: We should use the same approach for all slices.

src/main/java/com/artipie/http/ArtipieRepositories.java
```diff
@@ -75,21 +75,27 @@ public Slice slice(final Key name, final boolean standalone) {
...
+        final int port
     ) {
         return new RepositoriesFromStorage(this.http, storage).config(name.string()).thenCombine(
             StorageAliases.find(storage, name),
-            (config, aliases) -> new SliceFromConfig(
-                this.http, this.settings,
-                config, aliases, standalone
-            )
+            (config, aliases) -> {
+                Slice res = new SliceSimple(new RsRepoNotFound(name));
```
dgarus, line 90: Think it'll be better to use `else` brunch to create `SliceSimple`.

dgarus (changes requested):

olenagerasimova:
@dgarus all fixed and todos added. Please, take a look

dgarus (approved):
@olenagerasimova thanks!

Issue #1031, opened 2022-06-15
Validate artipie's repositories readme files content


dgarus:
Repository - adapter:
 - check header (should be as maven-adapter has)
 - link to [roadmap](https://github.com/orgs/artipie/projects/3) 
 - link to telegram chat: https://t.me/artipie
 - maven dependency (if adapter has)
 - description of repository structure and link to the repository documentation the implementation is based on
 - description of SDK and examples how to use
 - description how to configure a client to use the repository
 - `how to contribute` section is exist and contains `Please read [contributing rules](https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md)`
 -  add action for update license year

Artipie:
 - check header (should be as maven-adapter has)
 - link to telegram chat: https://t.me/artipie
 - link to [roadmap](https://github.com/orgs/artipie/projects/3)
 - check wiki content
 - how start
 - check all links
 - `how to contribute` section is exist and contains `Please read [contributing rules](https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md)`
 - add action for update license year

code examples for copy-paste is [here](https://github.com/artipie/artipie/issues/1031#issuecomment-1163026586)

Repositories:
- [ ] artipie
    - [x] main config
    - [ ] repos
    - [x] repo perms
    - [x] creds
    - [x] storages
    - [x] metrix
- [x] asto (+)
- [x] front (+)
- [x] conan-adapter (+)
- [x] http (+)
- [x] rpm-adapter (Alena) (+)
- [x] npm-adapter (Alena) (+)
- [x] gem-adapter (Denis) (+)
- [x] git-adapter
- [x] p2-adapter
- [x] go-adapter
- [x] files-adapter (Denis) (+)
- [x] docker-adapter (Denis) (+)
- [x] helm-adapter (Denis)
- [x] pypi-adapter (pypi-example) (+)
- [x] maven-adapter (+)
- [x] http-client (+)
- [x] vertx-server (+)
- [x] nuget-adapter
- [x] conda-adapter (+)
- [x] cargo-adapter
- [x] debian-adapter (+)
- [x] composer-adapter (Denis)
- [x] benchmarks

Swizbiz:
@artipie/members 
I think we should add this phrase to **How to contribute** 
> Please read [contributing rules](https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md).

Swizbiz:
@artipie/members 
I think we also have to bump year in license header to 2022

[...]

PR #1060, opened 2022-07-20
fix: correct examples status table and fix conda test

olenagerasimova:
Part of #1031 
Corrected examples status table, fixed pypi test, upgraded wiki sources.


dgarus (approved):
@olenagerasimova great, thanks!

PR #1062, opened 2022-07-21
doc: move all the configs details to wiki

olenagerasimova:
Part of #1031 

Created wiki sections structure and moved all the information from readme there. In readme I've left only general description, quickstart section and contributing details. 

Then I will extend and reformat each wiki page one by one.


dgarus (approved):
@olenagerasimova thanks!

Issue #1034, opened 2022-06-20
Ability to plug in a custom storage


dgarus:
Artipie should allow using a custom implementation of storage without code changes.
A custom storage must implement `Storage` interface and supplied in a separate jar file.

- [x] [Create a factory interface for an abstract storage](https://github.com/artipie/asto/issues/439)

PR #1086, opened 2022-08-02
feat: ability to plug in a custom storage

dgarus:
Closes: #1034

 - obtaining instance of storage using class `Storages`


olenagerasimova (approved):
@dgarus thanks

PR #1088, opened 2022-08-03
doc: restructured repo descriptions and added deb

olenagerasimova:
Part of #1031 
Restructured repositories descr files, corrected examples readme and removed not actual readmes from examples.


dgarus (approved):
thanks!

PR #1089, opened 2022-08-03
doc: corrected links and added another repos

olenagerasimova:
Part of #1031 
Corrected links (wiki does nott add subdirs into links, just md files names) and added more repo descriptions.


dgarus (approved):
Thanks!

Issue #1093, opened 2022-08-12
Remove `wtf.g4s8:tuples dependency`


dgarus:
Remove `wtf.g4s8:tuples dependency`, if necessary, use appropriate classes from common libraries.

PR #1094, opened 2022-08-12
Remove `wtf.g4s8:tuples`

dgarus:
Closes: #1093

- removed `wtf.g4s8:tuples` dependency
- moved dependencies with test scope to one place


olenagerasimova (approved):
@dgarus thanks

Issue #1099, opened 2022-08-16
Implement Rest API


olenagerasimova:
Let's implement Rest API to manage Artipie settings (repositories, storages, aliases, users). This Rest API will be a part of this project and will start on the separate port (another non-asynchronous http server). 
We'll use:
- spring boot
- swagger for docs (interfaces with docs and implementations with code)

Method description is available [here](https://github.com/artipie/front).

Take in account the issues #277, #127

dgarus:
@olenagerasimova @andpopov 

Related to https://github.com/artipie/artipie/pull/1119:
We have a lot of negative cases, but test only "duplicate repo" case.
Suggest add tests for following negative cases:
- Wrong repository name (one of reserved word).
- JSON body is expected.
- Section `repo` is required.
- Repository type is required.
- Repository storage is required.

olenagerasimova:
@dgarus the following cases
- JSON body is expected.
- Section repo is required.
- Repository type is required.
- Repository storage is required.

are handled by vertx and open api spec, `bad request` is automatically returned if body or any required field is missing.

I have another question: what auth type we will use fot API?

dgarus:
> I have another question: what auth type we will use fot API?

The same, which are supported by front currently.
If there is a proposal, then welcome.

olenagerasimova:
@dgarus on `front` we implemented JWT token auth, with vertx we can make it easily (as it has corresponding auth implementation). Authorization and permissions is another level, I think we should discuss and design it along with remaking all artipie permissions system.

PR #1102, opened 2022-08-19
feat: simplify `RepoConfig` class

olenagerasimova:
Part of #1099 
I've simplified `RepoConfig` class by:
- moving static method `fromPublisher()` to `RepositoriesFromStorage` class
- removing and inlining method `proxy()`


dgarus (approved):
@olenagerasimova thanks!

PR #1103, opened 2022-08-22
feat: added skeletons for repo configs operations

olenagerasimova:
Part of #1099 
Added interfaces for repository settings Rest API operations: to manage repos and repo permissions settings. 
Also, removed unused and not required any more classes, moved some classes from/to `repo` package.


dgarus (changes requested):
@olenagerasimova thanks!
Take a look please on my comments.

olenagerasimova:
@dgarus thanks
I've renamed interfaces and added DTO interfaces (I think they better be interfaces so we can make various implementations in the future and use them as for read as for write purposes).

olenagerasimova:
@dgarus please, take a look

dgarus (approved):
@olenagerasimova thanks!

PR #1108, opened 2022-08-25
feat: added test for RepositoryVerticle

olenagerasimova:
part of #1099 
Added test for RepositoryVerticle

### DIFF
--- a/src/main/java/com/artipie/api/RepositoryVerticle.java
+++ b/src/main/java/com/artipie/api/RepositoryVerticle.java
@@ -110,12 +110,13 @@ public final class RepositoryVerticle extends AbstractVerticle {
     }
 
     /**
-     * List all existing repositories.
+     * List user's repositories.
      * @param context Routing context
      */
     private void listUserRepos(final RoutingContext context) {
         context.response().setStatusCode(HttpStatus.OK_200).end(
-            JsonArray.of(this.crs.list(context.pathParam(RepositoryVerticle.UNAME))).encode()
+            JsonArray.of(this.crs.list(context.pathParam(RepositoryVerticle.UNAME)).toArray())
+                .encode()
         );
     }
 
--- a/src/test/java/com/artipie/api/RepositoryVerticleTest.java
+++ b/src/test/java/com/artipie/api/RepositoryVerticleTest.java
@@ -53,6 +53,7 @@ class RepositoryVerticleTest {
             ),
             ctx.succeedingThenComplete()
         );
+        // @checkstyle MagicNumberCheck (1 line)
         Thread.sleep(3000);
     }
 
--- a/src/test/java/com/artipie/api/RepositoryVerticleTest.java
+++ b/src/test/java/com/artipie/api/RepositoryVerticleTest.java
@@ -7,11 +7,13 @@ package com.artipie.api;
 import com.artipie.asto.Key;
 import com.artipie.asto.blocking.BlockingStorage;
 import com.artipie.asto.memory.InMemoryStorage;
+import com.artipie.nuget.RandomFreePort;
 import io.vertx.core.Vertx;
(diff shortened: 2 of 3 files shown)


src/test/java/com/artipie/api/RepositoryVerticleTest.java
```diff
@@ -0,0 +1,98 @@
...
+import io.vertx.core.http.HttpMethod;
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+import java.util.stream.Collectors;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
```
andpopov, line 21: @olenagerasimova Test is down on my pc, because HttpServer starting takes much time:
io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: no further information: localhost/127.0.0.1:8089

In my test I added such waiter: waitHttpServerUp(vertx)
https://github.com/andpopov/artipie/blob/integration_vertx-openapi/src/test/java/com/artipie/PetsopenapiTest.java

src/test/java/com/artipie/api/RepositoryVerticleTest.java
```diff
@@ -35,17 +37,18 @@ class RepositoryVerticleTest {
     /**
      * Service port.
      */
-    private static final int PORT = 8089;
+    private static int port;
```
andpopov, line 41: @olenagerasimova Test crashed again.
The cause of it is test starts before http server starts to listen incommig port.
It is required to wait http server port get available.

src/test/java/com/artipie/api/RepositoryVerticleTest.java
olenagerasimova: I've seen the waiter, it causes thread blocked error... Also, in my opinion good-old thread-wait is better as it does not add any complication to the test...

src/test/java/com/artipie/api/RepositoryVerticleTest.java
```diff
@@ -52,6 +53,7 @@ static void prepare(final Vertx vertx, final VertxTestContext ctx) throws IOExce
             ),
```
andpopov, line 53: Test works.
I would add ping of port during some time period, to avoid time consumption of test, because time period may be more or less than 3 seconds, it depends on several conditions. If test ping every half of second and period is shorter than 3 second then test will start quicker.

andpopov (approved):

src/main/java/com/artipie/api/RepositoryVerticle.java
```diff
@@ -110,12 +110,13 @@ private void listAll(final RoutingContext context) {
...
 
     /**
-     * List all existing repositories.
+     * List user's repositories.
      * @param context Routing context
      */
     private void listUserRepos(final RoutingContext context) {
         context.response().setStatusCode(HttpStatus.OK_200).end(
-            JsonArray.of(this.crs.list(context.pathParam(RepositoryVerticle.UNAME))).encode()
+            JsonArray.of(this.crs.list(context.pathParam(RepositoryVerticle.UNAME)).toArray())
```
dgarus, line 118: What about logging of requests/responses?
Do we already have it from the box?

src/test/java/com/artipie/api/RepositoryVerticleTest.java
andpopov: ping will allow to start test early than general sleep/wait.

andpopov (changes requested):

dgarus (approved):

andpopov (approved):

PR #1109, opened 2022-08-26
feat: add awaiting of server port availability for RepositoryVerticle…

andpopov:
add awaiting of server port availability for RepositoryVerticleTest

### DIFF
--- a/src/test/java/com/artipie/api/RepositoryVerticleTest.java
+++ b/src/test/java/com/artipie/api/RepositoryVerticleTest.java
@@ -11,12 +11,17 @@ import com.artipie.nuget.RandomFreePort;
 import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpClientResponse;
 import io.vertx.core.http.HttpMethod;
+import io.vertx.core.net.NetClient;
 import io.vertx.junit5.VertxExtension;
 import io.vertx.junit5.VertxTestContext;
 import java.io.IOException;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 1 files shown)


src/test/java/com/artipie/api/RepositoryVerticleTest.java
```diff
@@ -101,4 +124,35 @@ void listsUserRepos(final Vertx vertx, final VertxTestContext ctx) {
...
+            if (!available.get()) {
+                try {
+                    TimeUnit.MILLISECONDS.sleep(RepositoryVerticleTest.SLEEP_DURATION.toMillis());
+                } catch (final InterruptedException err) {
+                    break;
+                }
+            }
+        }
+        if (!available.get()) {
+            Assertions.fail();
```
dgarus, line 155: Maybe pass to `fail` a message which would explain what is wrong?

src/test/java/com/artipie/api/RepositoryVerticleTest.java
```diff
@@ -34,14 +39,33 @@ class RepositoryVerticleTest {
...
+
     /**
      * Service port.
      */
     private static int port;
 
+    /**
+     * Maximum awaiting time duration of port availability.
+     */
+    private static final Duration MAX_WAIT = Duration.ofMinutes(10);
```
dgarus, line 55: Is there any sense to use `Duration` rather than simple long?

src/test/java/com/artipie/api/RepositoryVerticleTest.java
```diff
@@ -34,14 +39,33 @@ class RepositoryVerticleTest {
...
+
+    /**
+     * Sleep duration.
+     */
+    private static final Duration SLEEP_DURATION = Duration.ofMillis(100);
```
dgarus, line 60: Is there any sense to use Duration rather than simple long?

dgarus (changes requested):
@andpopov thanks!
There are a few minor comments.

src/test/java/com/artipie/api/RepositoryVerticleTest.java
dgarus: Ok, I'm not against using `Duration`, but do we really need to convert every time `duration` to long?
We can do it once:
```
private static final long MAX_WAIT = Duration.ofMinutes(10).toMillis();
```

src/test/java/com/artipie/api/RepositoryVerticleTest.java
andpopov: @dgarus Thanks for good proposal.
I have added more informative message

src/test/java/com/artipie/api/RepositoryVerticleTest.java
andpopov: Fixed. Please review

src/test/java/com/artipie/api/RepositoryVerticleTest.java
```diff
@@ -34,14 +39,34 @@ class RepositoryVerticleTest {
...
+     * @checkstyle MagicNumberCheck (10 lines)
+     */
+    private static final long MAX_WAIT = Duration.ofMinutes(10).toMillis();
+
+    /**
+     * Sleep duration.
+     */
+    private static final long SLEEP_DURATION = Duration.ofMillis(100).toMillis();
+
+    static {
```
dgarus, line 63: Looks like over checking.
There is no needing to check constant value, it makes sense for function arguments.
I suggest removing these checking.

dgarus (changes requested):
Thanks!
Please have a look at my comment.

src/test/java/com/artipie/api/RepositoryVerticleTest.java
andpopov: removed.
@dgarus please review

dgarus (approved):
Thanks!

Issue #1110, opened 2022-08-29
feat: add rest-api for PUT repository


andpopov:
Part of #1099
Add openapi rest-implementation for PUT repository operation.

PR #1113, opened 2022-08-30
feat: add rest-api for PUT repository

andpopov:
#1110
Added PUT  implementation for repository
Dear @olenagerasimova and @dgarus Please review


src/main/java/com/artipie/settings/repo/CrudRepoSettings.java
```diff
@@ -28,27 +28,48 @@ public interface CrudRepoSettings {
 
     /**
      * Check if the repository exists.
-     * @param name Repository name. The name can be composite: in the case of org layout it will
-     *  consist of two parts - username and repo name, for example john/maven-s3.
+     * @param rname Repository name, for example maven-s3.
      * @return True if repository exists
      */
-    boolean exists(String name);
+    boolean exists(String rname);
```
olenagerasimova, line 34: I do not think this interface should know anything about combined (with username) or not repository names. This logic should be processed on the upper level, here already finalized repository name should be passed

src/main/java/com/artipie/api/ManageRepoSettings.java
```diff
@@ -6,20 +6,21 @@
...
-import javax.json.JsonStructure;
+import java.util.function.Function;
 import org.apache.commons.lang3.NotImplementedException;
 
 /**
  * Manage repository settings.
  * @since 0.26
  */
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.AvoidDuplicateLiterals"})
 public final class ManageRepoSettings implements CrudRepoSettings {
```
olenagerasimova, line 23: please, do not use `AvoidDuplicateLiterals` in the main code

src/main/java/com/artipie/api/ManageRepoSettings.java
```diff
@@ -50,18 +51,41 @@ public Collection<String> list(final String uname) {
...
-        throw ManageRepoSettings.NOT_IMPLEMENTED;
+    public boolean exists(final String rname) {
+        return this.exists(rname, ext -> ext.key(rname));
     }
 
     @Override
-    public JsonStructure value(final String name) {
-        return Json.createObjectBuilder().add("type", "maven").add("storage", "def").build();
+    public boolean exists(final String uname, final String rname) {
+        return this.exists(rname, ext -> ext.key(String.format("%s/%s", uname, rname)));
```
olenagerasimova, line 60: use `key` method here

src/main/java/com/artipie/api/ManageRepoSettings.java
```diff
@@ -50,18 +51,41 @@ public Collection<String> list(final String uname) {
...
     }
 
     @Override
-    public void save(final String name, final JsonStructure value) {
-        throw ManageRepoSettings.NOT_IMPLEMENTED;
+    public JsonObject value(final String rname) {
+        final byte[] value = this.asto.value(ManageRepoSettings.key(rname));
```
olenagerasimova, line 65: variable can be inlined, redundant variables are not welcome in Artipie 
https://www.yegor256.com/2015/09/01/redundant-variables-are-evil.html

src/main/java/com/artipie/api/ManageRepoSettings.java
```diff
@@ -50,18 +51,41 @@ public Collection<String> list(final String uname) {
...
+        return new JsonObject(new String(value, StandardCharsets.UTF_8));
+    }
+
+    @Override
+    public JsonObject value(final String uname, final String rname) {
+        final byte[] value = this.asto.value(ManageRepoSettings.key(uname, rname));
```
olenagerasimova, line 71: the same

olenagerasimova (changes requested):
@andpopov @dgarus I see two serious problems here:
- repository settings are stored in YAML formats, not json ([check wiki](https://github.com/artipie/artipie/wiki/Configuration-Repository))
- `org` and `flat` layouts are two different layouts, they cannot intersect and thus if we are working with `flat` layout we cannot have repos by `uname/rname` path, `uname` is not used. The same is true for `org` layout: repositories ALWAYS have username prefix. Also, we should show different rest api endpoints for different layouts.

src/main/java/com/artipie/api/RepositoryName.java
```diff
@@ -0,0 +1,75 @@
...
+ * https://github.com/artipie/artipie/LICENSE.txt
+ */
+
+package com.artipie.api;
+
+/**
+ * Repository name.
+ * @since 0.26
+ */
+public interface RepositoryName {
```
olenagerasimova, line 12: I'd suggest to add implementation `FromRequest` that will accept `RoutingContext` and layout into ctor and obtain repository name in `string()` method with the logic to obtain repo name from either one request parameter from two.

src/main/resources/swagger-ui/yaml/repository.yaml
```diff
@@ -1,9 +1,17 @@
 openapi: "3.0.0"
```
olenagerasimova, line 1: also, as we discussed: let's create separate yamles for org (requests with usernames) and flat (without usernames) layout, as users should not see requests they are not allowed to perform in swagger.

olenagerasimova (changes requested):
@andpopov thanks, added comments wee already discussed

src/main/resources/swagger-ui/yaml/repository.yaml
andpopov: @olenagerasimova, @dgarus Please review

src/main/java/com/artipie/api/RepositoryName.java
```diff
@@ -0,0 +1,122 @@
...
+            if ("flat".equals(layout)) {
+                this.origin = new RepositoryName.FlatRepositoryName(
+                    ctx.pathParam(RepositoryName.RNAME)
+                );
+            } else {
+                this.origin = new RepositoryName.OrgRepositoryName(
+                    ctx.pathParam(RepositoryName.UNAME),
+                    ctx.pathParam(RepositoryName.RNAME)
+                );
+            }
```
olenagerasimova, line 59: hm, I'd suggest putting all the logic inside `string()` method, then we will not disrupt `ConstructorOnlyInitializesOrCallOtherConstructors` rule and can get rid of implementations `FlatRepositoryName` and `OrgRepositoryName` as they do nothing...

src/main/java/com/artipie/api/RepositoryRest.java
```diff
@@ -0,0 +1,182 @@
...
+                .handler(this::createRepo)
+                .failureHandler(this.errorHandler(HttpStatus.INTERNAL_SERVER_ERROR_500));
+        }
+        final Router router = rbr.createRouter();
+        router.route("/api/*").handler(
+            StaticHandler.create(
+                FileSystemAccess.ROOT,
+                new JavaResource("swagger-ui").uri().getPath()
+            )
+        );
```
olenagerasimova, line 77: Do we need to expose swagger here? As far as I understand, we do in `RestApi` class with correct index

olenagerasimova (changes requested):
@andpopov thanks, check my comments please

src/main/java/com/artipie/api/RepositoryRest.java
andpopov: All content of swagger-ui folder is exposed outside including yaml-folder and index-org.html, index-flat.html

andpopov: RestApi is up the rest controller that build based on yaml open api spec.
But swagger-ui should be exposed because it shows yaml in swagger UI.

olenagerasimova (approved):
@andpopov thanks
@dgarus please, take a look

src/main/java/com/artipie/api/RepositoryName.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/artipie/LICENSE.txt
+ */
+
```
dgarus, line 5: extra line

src/main/java/com/artipie/api/RepositoryName.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    String RNAME = "rname";
+
+    /**
+     * String representation of repository name.
+     * For flat layout consists of 'reponame'
+     * For org layout consists of 'username/reponame'
+     * @return Repository name as string
+     */
+    String string();
```
dgarus, line 31: Suggest to rename this method to `toString` to override `Object` method.

src/main/java/com/artipie/api/ManageRepoSettings.java
```diff
@@ -84,13 +115,42 @@ private Collection<String> list(final Key key) {
...
+        return allowedReponame(rname.string());
+    }
+
+    /**
+     * Check if the repository name is allowed.
+     * @param name Repository name.
+     * @return True if repository name is allowed
+     */
+    private static boolean allowedReponame(final String name) {
+        return !name.contains("_storages")
```
dgarus, line 140: We can use static Set.
```
private static final Set<String> SERVICE_NAMES = Collections.unmodifiableSet(
        Sets.newHashSet("_storages", "_permissions", "_credentials")
    );
```
and 
```
return !SERVICE_NAMES.contains(name);
```

src/main/java/com/artipie/api/RepositoryRest.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Create a repository.
+     * @param context Routing context
+     */
+    private void createRepo(final RoutingContext context) {
+        final RepositoryName rname = new RepositoryName.FromRequest(context, this.layout);
+        if (this.crs.exists(rname)) {
```
dgarus, line 124: If repo name is a service name ("_storages", "_permissions", "_credentials") then this method returns false, 
and we will create repo with a service name.
It's wrong.

src/test/java/com/artipie/api/RepositoryRestTest.java
```diff
@@ -121,6 +129,72 @@ void listsUserRepos(final Vertx vertx, final VertxTestContext ctx) {
         );
     }
 
+    @Test
```
dgarus, line 132: We also should test negative cases.

src/main/java/com/artipie/api/RepositoryRest.java
```diff
@@ -0,0 +1,182 @@
...
+        );
+    }
+
+    /**
+     * Get a repository settings json.
+     * @param context Routing context
+     */
+    private void getRepo(final RoutingContext context) {
+        final RepositoryName rname = new RepositoryName.FromRequest(context, this.layout);
+        if (!this.crs.exists(rname)) {
```
dgarus, line 86: When repo name is a service name ("_storages", "_permissions", "_credentials"), we should give to user a proper error message. I think, NOT_FOUND is misleading in that case.

dgarus (changes requested):
@andpopov thanks!
There are comments, please, have a look.

src/main/java/com/artipie/api/RepositoryRest.java
andpopov: Thanks for your commets, please see change-commit.

olenagerasimova (approved):
@andpopov thanks
@dgarus I think we are good to merge here

dgarus (approved):
@andpopov thanks!

PR #1117, opened 2022-09-01
feat: crud for storage aliases

olenagerasimova:
Part of #1099 
Moved from front interface and implementation to manage storage aliases

### DIFF
--- a/src/main/java/com/artipie/api/ManageStorageAliases.java
+++ b/src/main/java/com/artipie/api/ManageStorageAliases.java
@@ -1,0 +1,216 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/artipie/LICENSE.txt
+ */
+package com.artipie.api;
+
+import com.amihaiemil.eoyaml.Yaml;
+import com.amihaiemil.eoyaml.YamlMapping;
+import com.amihaiemil.eoyaml.YamlMappingBuilder;
+import com.amihaiemil.eoyaml.YamlNode;
+import com.artipie.asto.Key;
+import com.artipie.asto.blocking.BlockingStorage;
+import com.artipie.misc.Json2Yaml;
+import com.artipie.misc.Yaml2Json;
+import com.artipie.settings.CrudStorageAliases;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import javax.json.JsonObject;
+
+/**
+ * Manage storage aliases settings.
+ * @since 0.27
+ */
+public final class ManageStorageAliases implements CrudStorageAliases {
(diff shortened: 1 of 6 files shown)


src/main/java/com/artipie/api/ManageStorageAliases.java
```diff
@@ -0,0 +1,216 @@
...
+
+    /**
+     * Yaml storages section name.
+     */
+    private static final String STORAGES_NODE = "storages";
+
+    /**
+     * Repository or user key.
+     */
+    private final Optional<Key> key;
```
dgarus, line 51: I don't like `Optional` as a field type or an argument type.
Maybe just `Key` and use `Optional.ofNullable(this.key)`  if `Optional` is necessary?

src/main/java/com/artipie/settings/CrudStorageAliases.java
```diff
@@ -0,0 +1,53 @@
...
+     * Remove storage from settings.
+     * @param alias Storage alias
+     */
+    void remove(String alias);
+
+    /**
+     * Artipie storage.
+     * @since 0.1
+     */
+    interface Storage {
```
dgarus, line 39: Class name `Storage` can be misleading because we already have `Storage` interface.

src/main/java/com/artipie/api/ManageStorageAliases.java
```diff
@@ -0,0 +1,216 @@
...
+     * @return The key if found
+     */
+    private Optional<Key> settingKey() {
+        Optional<Key> res = Optional.of(
+            this.key.<Key>map(val -> new Key.From(val, ManageStorageAliases.YAML))
+                .orElse(ManageStorageAliases.YAML)
+        );
+        if (!this.blsto.exists(res.get())) {
+            final String yml = String.format("%s.yml", ManageStorageAliases.FILE_NAME);
+            res = Optional.of(
```
dgarus, line 170: It looks like the same above

dgarus (changes requested):
@olenagerasimova thanks!
a few comments

src/main/java/com/artipie/api/ManageStorageAliases.java
olenagerasimova: I strongly disagree with nullable fields, storage aliases can be sett on different levels: for one repository, for user or for whole artipie. In the last case, the key will be empty.

src/main/java/com/artipie/api/ManageStorageAliases.java
olenagerasimova: it's not the same: above we `yaml` extension, here we use `yml`

olenagerasimova:
@dgarus thanks, answered, corrected and moved test from front

src/main/java/com/artipie/api/ManageStorageAliases.java
dgarus: Ok, I didn't notice an extension

olenagerasimova:
@dgarus added another constant, take a look please

PR #1125, opened 2022-09-07
feat: add/update aliases

olenagerasimova:
part of #1099
Added endpoints to add/update aliases


dgarus (approved):
@olenagerasimova thanks!

PR #1134, opened 2022-09-12
feat: added endpoint to get users list

olenagerasimova:
Part of #1099 
Added endpoint to get users list. 
As endpoint to manage users are the same for org and flat layouts, I've added new single open api yaml. In swagger, repositories and users APIs will be available by user's chose, check the screenshot:
image


andpopov (approved):
Looks good, thanks!

andpopov (approved):

PR #1135, opened 2022-09-13
feat: get user info

olenagerasimova:
Part of #1099 
Fixed `ManageUsers#list()` to remove user's password from user info, we should not return the password in rest api responses and added get method to obtain single user info.

### DIFF
--- a/src/main/java/com/artipie/api/UsersRest.java
+++ b/src/main/java/com/artipie/api/UsersRest.java
@@ -7,6 +7,8 @@ package com.artipie.api;
 import com.artipie.settings.users.CrudUsers;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.openapi.RouterBuilder;
+import java.util.Optional;
+import javax.json.JsonValue;
 import org.eclipse.jetty.http.HttpStatus;
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/com/artipie/api/UsersRest.java
```diff
@@ -33,15 +35,33 @@ public void init(final RouterBuilder rbr) {
...
+            .handler(this::getUser)
+            .failureHandler(this.errorHandler(HttpStatus.INTERNAL_SERVER_ERROR_500));
+    }
+
+    /**
+     * Get single user info.
+     * @param context Request context
+     */
+    private void getUser(final RoutingContext context) {
+        final Optional<JsonValue> usr = this.users.list().stream().filter(
```
andpopov, line 48: can we add method CrudUsers.user(final String name) and use it?
Because we convert all Yaml-users to JsonArray and afteward search in JsonArray, I seem it is memory consumption.

andpopov (changes requested):

olenagerasimova:
@andpopov thanks, added method

dgarus (approved):
@olenagerasimova thanks!

andpopov (approved):
@olenagerasimova thanks!

PR #1138, opened 2022-09-14
fix: Broken links on wiki pages #1137

andpopov:
close: #1137
@olenagerasimova @Swizbiz @dgarus could you please to review?

Issue #1137, opened 2022-09-14
Broken links on wiki pages


andpopov:
Now, there are many broken inner links on wiki 'repositories'-pages

PR #1139, opened 2022-09-14
fix: Broken links on wiki pages #1137

andpopov:
close: #1138 
Remove .md extension in links


olenagerasimova (approved):
@andpopov thanks

PR #1140, opened 2022-09-15
feat: delete user endpoint

olenagerasimova:
Part pf #1099 
Added endpoint to remove user

### DIFF
--- a/src/main/java/com/artipie/api/UsersRest.java
+++ b/src/main/java/com/artipie/api/UsersRest.java
@@ -6,6 +6,7 @@ package com.artipie.api;
 
 import com.artipie.settings.cache.AuthCache;
 import com.artipie.settings.users.CrudUsers;
+import com.jcabi.log.Logger;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.openapi.RouterBuilder;
 import java.io.StringReader;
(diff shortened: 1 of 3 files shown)


src/main/java/com/artipie/api/UsersRest.java
```diff
@@ -51,6 +52,24 @@ public void init(final RouterBuilder rbr) {
...
+    }
+
+    /**
+     * Removes user.
+     * @param context Request context
+     */
+    private void deleteUser(final RoutingContext context) {
+        try {
+            this.users.remove(context.pathParam(RepositoryName.UNAME));
+            this.cache.invalidateAll();
```
andpopov, line 67: Can we to narrow the 'try'-block to catch exactly exception rom 'users.remove(...)'?
I mean:
        try {
            this.users.remove(context.pathParam(RepositoryName.UNAME));
        } catch (final IllegalStateException err) {
            Logger.error(this, err.getMessage());
            context.response().setStatusCode(HttpStatus.NOT_FOUND_404).end();
        }
        this.cache.invalidateAll();
        context.response().setStatusCode(HttpStatus.OK_200).end();

andpopov (changes requested):

olenagerasimova:
@andpopov thanks, corrected

dgarus (approved):
@olenagerasimova thanks!

src/main/java/com/artipie/api/UsersRest.java
andpopov: @olenagerasimova thanks!

andpopov (approved):

PR #1149, opened 2022-09-23
feat: added more tests for authentication

olenagerasimova:
Part of #1099 
Added more tests for authentication.


dgarus (approved):
@olenagerasimova thanks!
LGTM

PR #1150, opened 2022-09-26
feat: added endpoint to change user password

olenagerasimova:
Part of #1099 

Added endpoint to change user password

### DIFF
--- a/src/main/java/com/artipie/api/ManageUsers.java
+++ b/src/main/java/com/artipie/api/ManageUsers.java
@@ -34,6 +34,16 @@ import javax.json.JsonObjectBuilder;
 public final class ManageUsers implements CrudUsers {
 
     /**
+     * Yaml field name for password.
+     */
+    private static final String PASS = "pass";
+
+    /**
+     * Yaml field name for password type.
+     */
+    private static final String TYPE = "type";
+
+    /**
      * Yaml file key.
      */
     private final Key key;
@@ -73,7 +83,7 @@ public final class ManageUsers implements CrudUsers {
     }
 
     @Override
-    public void addOrUpdate(final JsonObject info, final String uid) {
+    public void addOrUpdate(final JsonObject info, final String uname) {
         YamlMappingBuilder builder = Yaml.createYamlMappingBuilder();
         final Optional<YamlMapping> users = this.users();
         if (users.isPresent()) {
@@ -82,7 +92,7 @@ public final class ManageUsers implements CrudUsers {
                 builder = builder.add(val, users.get().yamlMapping(val));
             }
         }
-        builder = builder.add(uid, new Json2Yaml().apply(info.toString()));
+        builder = builder.add(uname, new Json2Yaml().apply(info.toString()));
         this.blsto.save(
             this.key,
             Yaml.createYamlMappingBuilder().add(YamlSettings.NODE_CREDENTIALS, builder.build())
@@ -91,13 +101,13 @@ public final class ManageUsers implements CrudUsers {
     }
 
     @Override
-    public void remove(final String uid) {
-        if (this.users().map(yaml -> yaml.yamlMapping(uid) != null).orElse(false)) {
+    public void remove(final String uname) {
+        if (this.users().map(yaml -> yaml.yamlMapping(uname) != null).orElse(false)) {
             YamlMappingBuilder builder = Yaml.createYamlMappingBuilder();
             final YamlMapping users = this.users().get();
             for (final YamlNode node : users.keys()) {
                 final String val = node.asScalar().value();
-                if (!uid.equals(val)) {
+                if (!uname.equals(val)) {
                     builder = builder.add(val, users.yamlMapping(val));
                 }
             }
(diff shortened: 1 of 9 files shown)


src/main/java/com/artipie/api/ManageUsers.java
```diff
@@ -109,7 +119,40 @@ public void remove(final String uid) {
...
-        throw new IllegalStateException(String.format("User %s does not exist", uid));
+        throw new IllegalStateException(String.format("User %s does not exist", uname));
+    }
+
+    @Override
+    public void alterPassword(final String uname, final JsonObject info) {
+        if (this.users().map(yaml -> yaml.yamlMapping(uname) != null).orElse(false)) {
+            YamlMappingBuilder all = Yaml.createYamlMappingBuilder();
+            final YamlMapping users = this.users().get();
+            for (final YamlNode node : users.keys()) {
```
andpopov, line 130: Lines 127 and 129 invoke the same method: this.users()
I would to store result and reuse it.

src/main/java/com/artipie/api/ManageUsers.java
```diff
@@ -109,7 +119,40 @@ public void remove(final String uid) {
...
+                final String val = node.asScalar().value();
+                if (!uname.equals(val)) {
```
andpopov, line 132: Note no change required: It is just example of usage standard java style of coding, it is not requirement to modify code:
Line 131: for variable 'val' is good name is 'userName'
It is good example for usage CamelCase

src/main/java/com/artipie/api/ManageUsers.java
```diff
@@ -109,7 +119,40 @@ public void remove(final String uid) {
...
+                    all = all.add(val, users.yamlMapping(val));
+                }
+            }
+            final YamlMapping user = this.users().get().yamlMapping(uname);
+            YamlMappingBuilder changing = Yaml.createYamlMappingBuilder();
```
andpopov, line 137: Line 136 is the same as : 127 and 129

src/main/java/com/artipie/api/ManageUsers.java
```diff
@@ -109,7 +119,40 @@ public void remove(final String uid) {
...
+            for (final YamlNode node : user.keys()) {
+                final String val = node.asScalar().value();
+                if (!ManageUsers.TYPE.equals(val) && !ManageUsers.PASS.equals(val)) {
```
andpopov, line 140: Just information, changing is not mandatory.
Line 139: variable name 'val' is not informative, for example it may be 'key' or 'property'

src/main/java/com/artipie/api/ManageUsers.java
```diff
@@ -109,7 +119,40 @@ public void remove(final String uid) {
...
+            all = all.add(uname, changing.build());
+            this.blsto.save(
+                this.key,
+                Yaml.createYamlMappingBuilder()
+                    .add(YamlSettings.NODE_CREDENTIALS, all.build())
+                    .build().toString().getBytes(StandardCharsets.UTF_8)
+            );
+            return;
+        }
+        throw new IllegalStateException(String.format("User %s does not found", uname));
```
andpopov, line 155: Cannot find usage of 'old_pass'

src/test/java/com/artipie/api/ManageUsersTest.java
```diff
@@ -162,6 +162,39 @@ void removesUser() {
         );
     }
 
+    @Test
+    void altersPassword() {
+        this.blsto.save(
```
andpopov, line 167: 'old_pass' not found usage

andpopov (dismissed):

src/test/java/com/artipie/api/ManageUsersTest.java
olenagerasimova: check `UsersRest.java:128`

src/main/java/com/artipie/api/ManageUsers.java
olenagerasimova: check `UsersRest.java:128`, it is not a responsibility of `ManageUsers` class to verify current password, `UsersRest` does it before calling `ManageUsers.alterPassword()` to apply the changes

src/main/java/com/artipie/api/ManageUsers.java
olenagerasimova: `name` is enough in this case, we work with users here, so it's obvious that `name` is the name of a user

olenagerasimova:
@andpopov thanks, corrected and answered

andpopov (approved):
@olenagerasimova Thanks!

src/test/java/com/artipie/api/ManageUsersTest.java
```diff
@@ -162,6 +162,39 @@ void removesUser() {
         );
     }
 
+    @Test
+    void altersPassword() {
```
dgarus, line 166: Think we should have a test that checks the case when the old password is invalid.

dgarus (changes requested):
@olenagerasimova thanks!
We have to add tests for negative cases 404 and 401

dgarus (approved):
@olenagerasimova Thanks!
LGTM

PR #1151, opened 2022-09-26
feat: rest api port parameter

olenagerasimova:
Part of #1099 
Added a separate cmd parameter to specify port to start Artipie rest api on.

But in the case SSL is used, it seems we have to use port 443 and ignore this option...

### DIFF
--- a/src/main/java/com/artipie/VertxMain.java
+++ b/src/main/java/com/artipie/VertxMain.java
@@ -51,6 +51,11 @@ import org.apache.commons.cli.Options;
 public final class VertxMain {
 
     /**
+     * Default port to start Artipie Rest API service.
+     */
+    private static final String DEF_API_PORT = "8086";
+
+    /**
      * HTTP client.
      */
     private final ClientSlices http;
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/VertxMain.java
```diff
@@ -143,8 +148,10 @@ public static void main(final String... args) throws Exception {
         final Options options = new Options();
         final String popt = "p";
         final String fopt = "f";
-        options.addOption(popt, "port", true, "The port to start artipie on");
-        options.addOption(fopt, "config-file", true, "The path to artipie configuration file");
+        final String apiport = "apiport";
```
dgarus, line 151: "apiport" -> "api-port"

dgarus (changes requested):

src/main/java/com/artipie/VertxMain.java
olenagerasimova: @dgarus cmd option name cannot have dashes

dgarus: @olenagerasimova ok

dgarus (approved):
thanks!

andpopov (approved):
@olenagerasimova Looks good! Thanks

PR #1154, opened 2022-09-27
wiki: corrected repo table links

olenagerasimova:
Removed `.md` extensions from repository types table as with this extension links do not work correctly.


dgarus (approved):
thanks!

PR #1155, opened 2022-09-28
wiki: fix links

olenagerasimova:
Fixing wiki links again: they should not have `.md` at the end. Also, in the repository types table they do work, trying different variants.


dgarus (approved):
@olenagerasimova thanks!

Issue #1159, opened 2022-09-29
Add SSL support for REST API


andpopov:
Part of #1099
REST services API should be able to serve HTTP-clients in secure mode with enabled SSL.
It should be configurable through artipie.yaml so HTTP REST port can work either in HTTP or HTTPS.

PR #1158, opened 2022-09-29
wiki: added Rest API docs

olenagerasimova:
Part of #1099 

Added wiki page about rest api

PR #1160, opened 2022-09-29
Add SSL support for REST API

andpopov:
closes: #1159 

To configure SSL for REST-port the section 'ssl' should be added to artipie.yaml:
meta:
  ssl:

To switch SSL the 'ssl.enabled=true' should be set:
meta:
  ssl:
    enabled:true

One of types of SSL-keystores should be added: 'jks' or 'pem' or 'pfx'.

There are tests for each :
1. JKS - SSLJksRestTest
2. PEM - SSLPemRestTest
3. PFX - SSLPfxRestTest

There is README with description how keys can be created using keytool and openssl.

Examples of 3 configuration types:
meta:
  ssl:
    enabled:true
    pem:
      key-path:private-key.pem
      cert-path:cert.pem

meta:
  ssl:
    enabled:true
    jks:
      path:keystore.jks
      password:secret

meta:
  ssl:
    enabled:true
    pfx:
      path:cert.pfx
      password:secret

@olenagerasimova, @dgarus please review


src/main/java/com/artipie/api/KeyStore.java
```diff
@@ -0,0 +1,338 @@
...
+    @SuppressWarnings("PMD.ProhibitPublicStaticMethods")
+    public static KeyStore create(final YamlMapping yaml) {
+        final List<KeyStore> keystores = List.of(
+            new JksKeyStore(yaml), new PemKeyStore(yaml), new PfxKeyStore(yaml)
+        );
+        for (final KeyStore keystore : keystores) {
+            if (keystore.isConfigured()) {
+                return keystore;
+            }
+        }
```
olenagerasimova, line 85: hm, so if there are several key store types configured, we simply use any of them... I'm not sure, that's the best course of actions, as it's not obvious to user how we choose what type to use...

olenagerasimova (changes requested):
@andpopov thanks
@dgarus please, take a look at my comment about key store types, what do you think?

dgarus:
@andpopov thanks!
It would be great if you add to PR description what were done.
See our [contribution rules](https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md#description).

src/main/java/com/artipie/api/KeyStore.java
andpopov: @olenagerasimova if user specifies several will be used only one, we need to document usage SSL with examples and explanation.
I think that we met pitfail of simplicity of YAML because YAML does not support validation if we compare it with XML/XSD-schemas where XSDchema allows to restrict by only one of alternatives. May YAML-has some validation it would be good for us to validate artipie-configurations before storing them.

olenagerasimova: Documentation sure should be added in the next step, when this [PR](https://github.com/artipie/artipie/pull/1158) is merged.

src/main/java/com/artipie/api/KeyStore.java
```diff
@@ -0,0 +1,338 @@
...
+         * @param storage Storage.
+         * @return JKS-options for http server.
+         */
+        private JksOptions jksOptions(final Storage storage) {
+            if (!hasProperty(this.yaml(), JksKeyStore.JKS)) {
+                throw new IllegalStateException("'jks'-section is expected in yaml-configuration");
+            }
+            final YamlMapping jks = node(this.yaml(), JksKeyStore.JKS);
+            final JksOptions options = new JksOptions();
+            if (hasProperty(jks, KeyStore.PASSWORD)) {
```
dgarus, line 203: To reduce code duplication.
```
private static void setIfExists(YamlMapping mapping, String prop, Consumer<String> consumer) {
  if (hasProperty(mapping, prop)) {
     consumer.accept(mapping.string(prop));
  }
}
```

```
KeyStore.setIfExists(jks, KeyStore.PASSWORD, options::setPassword);
```

src/main/java/com/artipie/api/KeyStore.java
```diff
@@ -0,0 +1,338 @@
...
+        /**
+         * Initialize PFX-options based on yaml-configuration.
+         * @param storage Storage.
+         * @return PFX-options for http server.
+         */
+        private PfxOptions pfxOptions(final Storage storage) {
+            if (!hasProperty(this.yaml(), PfxKeyStore.PFX)) {
+                throw new IllegalStateException("'pfx'-section is expected in yaml-configuration");
+            }
+            final YamlMapping pfx = node(this.yaml(), PfxKeyStore.PFX);
```
dgarus, line 321: To reduce code duplications
```
private static KeyStoreOptionsBase options(
        YamlMapping mapping, 
        Supplier<KeyStoreOptionsBase> supplier) {
        KeyStoreOptionsBase res = supplier.get();
        KeyStore.setIfExists(mapping, KeyStore.PASSWORD, res::setPassword);
        KeyStore.setIfExists(mapping, KeyStore.ALIAS, res::setAlias);
        // ... 
        return res;
    }
```

src/main/java/com/artipie/api/KeyStore.java
```diff
@@ -0,0 +1,338 @@
...
+import io.vertx.core.net.JksOptions;
+import io.vertx.core.net.PemKeyCertOptions;
+import io.vertx.core.net.PfxOptions;
+import java.util.List;
+
+/**
+ * Key store.
+ * @since 0.26
+ */
+public abstract class KeyStore {
```
dgarus, line 24: Think we should create an interface `KeyStore` and abstract class `YamlBasedKeyStore`
and so on.

src/main/java/com/artipie/api/KeyStore.java
```diff
@@ -0,0 +1,338 @@
...
+     */
+    public abstract boolean isConfigured();
+
+    /**
+     * Create http server.
+     * @param vertx Vertx.
+     * @param storage Artipie settings storage.
+     * @return HttpServer
+     */
+    public abstract HttpServer createHttpServer(Vertx vertx, Storage storage);
```
dgarus, line 101: It looks strange that `KeyStore` creates `HttpServer`.
Suggest something like this:
```
public void updateOptions(
            final HttpServerOptions options,
            final Storage storage) {
            options.setPemKeyCertOptions(this.pemOptions(storage));
        }
```

src/test/java/com/artipie/api/RestApiServerBase.java
```diff
@@ -94,6 +97,27 @@ Authentication auth() {
...
+    }
+
+    /**
+     * Before each method initializes all necessary things that required for used KeyStore.
+     * @checkstyle NonStaticMethodCheck (5 lines)
+     */
+    @SuppressWarnings(
+        {"PMD.EmptyMethodInAbstractClassShouldBeAbstract", "PMD.UncommentedEmptyMethodBody"}
+    )
+    void initKeyStore() {
```
dgarus, line 118: It's an extra method.

src/test/java/com/artipie/api/RestApiServerBase.java
```diff
@@ -140,10 +164,14 @@ final void beforeEach(final Vertx vertx, final VertxTestContext context) throws
         final InMemoryStorage storage = new InMemoryStorage();
         this.asto = new BlockingStorage(storage);
         this.caches = new SettingsCaches.Fake();
+        if (this.keyStore().isPresent()) {
+            this.initKeyStore();
```
dgarus, line 168: You can do an initialization routine inside `keyStore()`.

dgarus (changes requested):
@andpopov thanks!
Please look at my comments.

src/main/java/com/artipie/settings/Settings.java
```diff
@@ -95,6 +102,11 @@ final class Fake implements Settings {
          */
         private final Layout layout;
 
+        /**
+         * KeyStore.
+         */
+        private final KeyStore keystore;
```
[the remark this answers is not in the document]
dgarus: I don't like `Optional` as a field type or as a method parameter.
`Optional` is not a silver bullet to avoid nulls, `Optional` reference itself can be `null`.

I like the `NullObject` pattern:
Everything that can be `null` has a null implementation.
For example, `EmptyKeyStore` that does nothing. So field cannot be `null`, it should be checked at initialization stage.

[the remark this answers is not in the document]
olenagerasimova: I like the direction, but it seem to be an overhead for this case: this `Fake` implementation can simply throw `NotImplementedException` for keystore-related methods as they are actually not used anywhere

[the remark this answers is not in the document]
dgarus: Sorry, I didn't understand your thought.
I'd like to note—it's appliable, if implement this [suggestion](https://github.com/artipie/artipie/pull/1160#discussion_r984392259).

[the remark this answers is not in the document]
olenagerasimova: `Fake` implementation of `Settings` is used for tests only and new keystore related methods are not actually called from any place of code. So we can simply throw `NotImplementedException` in this `Fake` implementation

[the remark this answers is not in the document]
dgarus: `NullObject` is not a fake or stub implementation.
`NullObject` is one of regular implementations to use in code base.

src/main/java/com/artipie/api/KeyStore.java
andpopov: Vertx allows to pass options only creation time of HttpServer

src/main/java/com/artipie/settings/Settings.java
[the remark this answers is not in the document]
olenagerasimova: I'm not saying that, check the class or the place (it's `Fake` implementation of `Settings` interface) we are talking about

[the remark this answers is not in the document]
olenagerasimova: `private final KeyStore keystore` is a field of `Fake` implementation of `Settings` interface, we can avoid adding this field at all, because it actually never used (key-store related methods of `Fake` implementation are never called)

src/main/java/com/artipie/api/KeyStore.java
dgarus: > Vertx allows to pass options only creation time of HttpServer

Ok, how does it influence on my comment?

src/main/java/com/artipie/settings/Settings.java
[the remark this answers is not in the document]
dgarus: Do you mean this field should be removed at all?

src/test/java/com/artipie/api/RestApiServerBase.java
andpopov: @olenagerasimova @dgarus Thank you for advices, please look at changes.

src/main/java/com/artipie/api/KeyStore.java
andpopov: I got your point, added enchancement.

src/main/java/com/artipie/settings/Settings.java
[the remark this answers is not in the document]
andpopov: For now it has Optional type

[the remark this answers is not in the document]
andpopov: For now it has Optional type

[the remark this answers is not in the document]
olenagerasimova: yes, the field can be removed and we can throw `NotImplementedException` inside key-store related methods

dgarus (approved):
Thanks!
LGTM

PR #1161, opened 2022-09-30
Optional '--jmx' for debug/test purposes

ChGen:
I suggest adding optional `--jmx` flag for debug/test purposes. It will be disabled by default. Usage:
`docker run .... -p 8888:8888 artipie/artipie --jmx`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ WORKDIR /var/artipie
 EXPOSE 8080 8086
 CMD [ \
   "java", \
+  $JVM_ARGS, \
   "--add-opens", "java.base/java.util=ALL-UNNAMED", \
   "--add-opens", "java.base/java.security=ALL-UNNAMED", \
   "-cp", "/usr/lib/artipie/artipie.jar:/usr/lib/artipie/lib/*", \
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,14 +18,5 @@ COPY target/${JAR_FILE} /usr/lib/artipie/artipie.jar
 VOLUME /var/artipie /etc/artipie
 WORKDIR /var/artipie
 EXPOSE 8080 8086
-CMD [ \
-  "java", \
-  $JVM_ARGS, \
-  "--add-opens", "java.base/java.util=ALL-UNNAMED", \
-  "--add-opens", "java.base/java.security=ALL-UNNAMED", \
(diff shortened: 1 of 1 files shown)


ChGen:

olenagerasimova:
@ChGen thanks, in my opinion it's more convenient and easier to read, when run command is inside docker file... I guess, it would be better to discuss with @dgarus

ChGen:
@dgarus , please review. Alena suggests moving launch back to the Dockerfile.

dgarus:
@ChGen 
How do you plane use it?

ChGen:
@dgarus , the idea is being able to to connect tools to published artipie docker images, so we can test or calculate statistics for our official releases.

dgarus:
> @dgarus , the idea is being able to to connect tools to published artipie docker images, so we can test or calculate statistics for our official releases.

Can we use environment variables to pass JVM options when running the container?

https://stackoverflow.com/questions/30494050/how-do-i-pass-environment-variables-to-docker-containers
https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file

ChGen:
@dgarus , yes, this could be done with something like `-e JMX_PORT=8888` instead. So if it's set we enable all JMX settings with specified port number.

dgarus:
@ChGen 
IMHO, hard-code for JVM options isn't a good solution 
because we may be needing to change them time to time.

ChGen:
@dgarus , yes, we can use `--env-file`  when starting our containers too, since it uses the same internal logic for passing environment inside container.

ChGen:
@dgarus , please review, I changed it to env. implementation.

launch.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+extra=""
+if [ -n "$JMX_PORT" ] ; then
+  extra="\
+  -Djava.rmi.server.hostname=localhost \
+  -Dcom.sun.management.jmxremote \
+  -Dcom.sun.management.jmxremote.port=$JMX_PORT \
```
dgarus, line 8: Can we pass all extra parameters as env-file?

dgarus (changes requested):
Thanks!
Please, look at my comment.

launch.sh
ChGen: Done

dgarus (changes requested):
Thanks!
Let's make without `launch.sh`

dgarus (approved):
thanks!

PR #1168, opened 2022-10-20
fix: AuthFromEnv corrected, hamcrest added to the main code

olenagerasimova:
`AuthFromEnv` should return empty optional if it fails to authorize user and `hamcrest` is required in the main code for `jcabi-github` to work properly.

Had to correct several tests and add credentials settings for smoke tests: `conda` client does not work without users and authorization at all, before the fix, Artipie simply used this anonymous from `AuthFromEnv`, now it uses user from credentials as should.

### DIFF
--- a/examples/artipie.yml
+++ b/examples/artipie.yml
@@ -3,3 +3,6 @@ meta:
   storage:
     type: fs
     path: /var/artipie/cfg
+  credentials:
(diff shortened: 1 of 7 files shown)


ChGen (approved):
@olenagerasimova , LGTM, thanks!

PR #1171, opened 2022-10-24
fix: use cached auth properly

olenagerasimova:
While reviewing https://github.com/artipie/front/pull/106 I run to guava cache "Recursive load" error in the case of specific auth configuration. The error was caused by double caching of github auth, which should not be done. Removed extra caching and corresponding classes, now all the caching is implemented via `SettingsCaches.All` class.


dgarus (approved):
@olenagerasimova thanks!
LGTM

PR #1173, opened 2022-11-02
feat: Getting layout setting via rest api

andpopov:

### DIFF
--- a/src/main/java/com/artipie/VertxMain.java
+++ b/src/main/java/com/artipie/VertxMain.java
@@ -115,7 +115,7 @@ public final class VertxMain {
             auth -> vertx.deployVerticle(
                 new RestApi(
                     caches, settings.repoConfigsStorage(), settings.layout().toString(),
-                    apiport, settings.credentialsKey(), auth, settings.keyStore()
+                    apiport, settings.credentialsKey(), auth, settings.keyStore(), settings
                 )
             )
         );
--- a/src/main/java/com/artipie/VertxMain.java
+++ b/src/main/java/com/artipie/VertxMain.java
@@ -115,7 +115,7 @@ public final class VertxMain {
             auth -> vertx.deployVerticle(
                 new RestApi(
                     caches, settings.repoConfigsStorage(), settings.layout().toString(),
-                    apiport, settings.credentialsKey(), auth, settings.keyStore(), settings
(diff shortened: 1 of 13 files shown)


src/main/java/com/artipie/api/SettingsRest.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * REST API methods to manage Artipie settings.
+ * @since 0.27
+ */
+public final class SettingsRest extends BaseRest {
+    /**
+     * Settings.
+     */
+    private final Settings settings;
```
olenagerasimova, line 22: Why do we have `Settings` here if the only thing we is layout?

src/main/java/com/artipie/api/RestApi.java
```diff
@@ -71,6 +72,11 @@ public final class RestApi extends AbstractVerticle {
      */
     private final Optional<KeyStore> keystore;
 
+    /**
+     * Artipie settings.
+     */
+    private final Settings settings;
```
olenagerasimova, line 78: I do not think we should have `Settings` as a field here, as it's not really required, layout is already in the class fields

src/main/java/com/artipie/VertxMain.java
```diff
@@ -115,7 +115,7 @@ public int start(final int apiport) throws IOException {
             auth -> vertx.deployVerticle(
                 new RestApi(
                     caches, settings.repoConfigsStorage(), settings.layout().toString(),
-                    apiport, settings.credentialsKey(), auth, settings.keyStore()
+                    apiport, settings.credentialsKey(), auth, settings.keyStore(), settings
```
olenagerasimova, line 118: and this looks strange, as we first ask credentials key and key store from the settings and then pass the whole settings instance...

olenagerasimova (changes requested):
@andpopov thanks, I think we can simply pass layout to `SettingsRest` instead of whole `Settings` instance

src/main/java/com/artipie/VertxMain.java
andpopov: @olenagerasimova Thanks, I have fixed, please review

olenagerasimova (approved):
@andpopov thanks

Issue #1175, opened 2022-11-03
Prometheus's metrics for Vertx and Artipie should use a single server


dgarus:
Currently, Prometheus's metrics for Vertx and Artipie are exposed by different servers. But it would be more convenient for clients to use one server to get all metrics. We should expose Prometheus's metrics for Vertx and Artipie using a single server.

[binding metrics to an existing vert x web router](https://vertx.io/docs/vertx-micrometer-metrics/java/#_binding_metrics_to_an_existing_vert_x_web_router)

PR #1176, opened 2022-11-07
Do not create metrics related slices when metrics are not defined

### LINKED ISSUE #1174: Create `TrafficMetricSlice`, `ResponseMetricsSlice` when according metric's settings require it
`BaseSlice` always creates `TrafficMetricSlice`, `ResponseMetricsSlice` regardless of  `measure` setting.
These classes add overhead to request handling.
We should create classes related to metrics only when according metric's settings require it.

dgarus:
Closes #1174

- Do not create metrics related slices when metrics are not defined
- Added exception when `NopMetrics` is used to publish metrics
- `PromuXXX` rename to `PrometheusXXX`
- `InMemoryMetrics` refactoring

### DIFF
--- a/src/main/java/com/artipie/http/MainSlice.java
+++ b/src/main/java/com/artipie/http/MainSlice.java
@@ -6,6 +6,7 @@ package com.artipie.http;
 
 import com.amihaiemil.eoyaml.YamlMapping;
 import com.amihaiemil.eoyaml.YamlNode;
+import com.amihaiemil.eoyaml.YamlSequence;
 import com.artipie.asto.Key;
 import com.artipie.asto.SubStorage;
 import com.artipie.http.client.ClientSlices;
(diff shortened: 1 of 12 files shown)


src/main/java/com/artipie/http/MainSlice.java
```diff
@@ -123,19 +123,16 @@ private static Optional<YamlMapping> metricsStorage(final Settings settings) {
...
-        final boolean available;
-        if (myml.isPresent()) {
-            final String type = myml.get().string("type");
-            available = MetricsFromConfig.PROMETHEUS.equals(type);
-        } else {
-            available = false;
+        boolean res = false;
+        final YamlMapping metrics = settings.meta().yamlMapping("metrics");
+        if (metrics != null) {
+            res = MetricsFromConfig.PROMETHEUS.equals(metrics.string("type"));
```
olenagerasimova, line 134: hm, this is actually my mistake, I forgot to fix this part when make `metrics` section yaml sequence instead of yaml mapping. Please, check metrics settings descriptions in [wiki](https://github.com/artipie/artipie/wiki/Configuration-Metrics) and let's correct this method accordingly...

olenagerasimova (changes requested):
@dgarus thanks, I've added one comment, please, take a look

src/main/java/com/artipie/http/MainSlice.java
```diff
@@ -112,30 +114,29 @@ public MainSlice(
...
-     * @return Yaml node, could be null
+     * @return True if metrics are collected by Prometheus.
      */
-    private static Optional<YamlMapping> metricsStorage(final Settings settings) {
-        return Optional.ofNullable(settings.meta().yamlMapping("metrics"))
-            .flatMap(metrics -> Optional.ofNullable(metrics.yamlMapping("storage")));
+    static boolean isPrometheusConfigAvailable(final Settings settings) {
+        return settings
+            .meta()
+            .yamlSequence("metrics")
```
olenagerasimova, line 125: if `metrics` section is absent at all, `yamlSequence("metrics")` returns null, and we will get NPE here

src/main/java/com/artipie/http/MainSlice.java
```diff
@@ -112,30 +114,29 @@ public MainSlice(
...
-        final boolean available;
-        if (myml.isPresent()) {
-            final String type = myml.get().string("type");
-            available = MetricsFromConfig.PROMETHEUS.equals(type);
-        } else {
-            available = false;
-        }
-        return available;
+    private static Optional<YamlMapping> metricsStorage(final Settings settings) {
+        return Optional.ofNullable(settings.meta().yamlMapping("metrics"))
```
olenagerasimova, line 139: Once again, `metrics` is not a mapping, but a sequence. In this sequence item with type `asto` should be found and on this item `storage` section is present...

olenagerasimova (changes requested):
@dgarus please, check my comments. Also, please, consider moving `metricsStorage` and `isPrometheusConfigAvailable` methods to `MetricsFromConfig` class. It seems these methods belongs there and if moved, they can be properly tested.

dgarus:
> @dgarus please, check my comments. Also, please, consider moving `metricsStorage` and `isPrometheusConfigAvailable` methods to `MetricsFromConfig` class. It seems these methods belongs there and if moved, they can be properly tested.

There is an issue #1175. So, these methods will moved later.

olenagerasimova (approved):
thanks

Issue #1177, opened 2022-11-07
Removing repostitory via REST-api should remove repository with wrong configuration


andpopov:
Currently repository configuration cannot be removed via REST-api if repository has incorrect storage configuration.

PR #1178, opened 2022-11-07
fix: Removing repostitory via REST-api should remove repository with wrong configuration

andpopov:
Closes: #1177


olenagerasimova (changes requested):
@andpopov looks reasonable, but what about test for this case?

andpopov:
@olenagerasimova Thanks, I have added tests, please review

olenagerasimova (approved):
@andpopov thanks

Issue #1179, opened 2022-11-08
Rework metrics related classes


dgarus:
Rework metrics related classes.
- all classes should be created only once
- executors should be shutdowned when the application is closing

PR #1181, opened 2022-11-11
Metrics context

dgarus:
closes: #1179

- added `MetricsContext` which replaces `MetricsFromConfig`
- `MetricsPublisher` encapsulated inside `MetricsContext` and has method to shut down executors
- added checks for metrics configuration


src/main/java/com/artipie/metrics/MetricsContext.java
```diff
@@ -0,0 +1,300 @@
...
+                                break;
+                            default:
+                                throw new IllegalStateException(
+                                    String.format(
+                                        "%s is not interval type of metrics output",
+                                        entry.getValue()
+                                    )
+                                );
+                        }
+                        return res;
```
olenagerasimova, line 142: I'd suggest simplifying this part by using enhanced switch (follow Idea's suggestions, it will bring you there:))

src/main/java/com/artipie/metrics/MetricsContext.java
```diff
@@ -0,0 +1,300 @@
...
+     * Port to expose vertx metrics.
+     *
+     * @return Port, if -1 then vertx metrics are not enabled.
+     */
+    public int vertxPort() {
+        int res = -1;
+        if (this.vertxEnabled()) {
+            res = Integer.parseInt(
+                this.outputs.get(MetricsOutputType.VERTX).string("port")
+            );
```
olenagerasimova, line 231: Suggesting to replace three methods with one, which will return `Optional<Pair<String, Integer>>` as these three methods are not consistent: calling `vertxPort()` does not make sense if such metric is disabled, `vertxEndpoint()` returns `null` with is not good and etc.

src/main/java/com/artipie/metrics/MetricsContext.java
```diff
@@ -0,0 +1,300 @@
...
+    public boolean enabled() {
+        return !this.outputs.isEmpty();
+    }
+
+    /**
+     * Log metrics output is enabled.
+     *
+     * @return True, if enabled.
+     */
+    public boolean logEnabled() {
```
olenagerasimova, line 163: this method is only used in tests

src/main/java/com/artipie/VertxMain.java
```diff
@@ -259,25 +246,28 @@ private static void logRepo(final int mport, final String name) {
...
-                        new MicrometerMetricsOptions()
-                            .setPrometheusOptions(
-                                new VertxPrometheusOptions().setEnabled(true)
-                                    .setStartEmbeddedServer(true)
-                                    .setEmbeddedServerOptions(
-                                        new HttpServerOptions().setPort(pair.getValue())
-                                    ).setEmbeddedServerEndpoint(pair.getKey())
-                            ).setEnabled(true)
-                    )
+    private static Vertx vertx(final MetricsContext mctx) {
```
olenagerasimova, line 253: as `MetricsContext` is a field now, I'd suggest to make this method not static

olenagerasimova (changes requested):
@dgarus thanks, consider my suggestions)

src/main/java/com/artipie/metrics/MetricsContext.java
dgarus: Unfortunately, `Qulice` doesn't know about enhanced switch and falls down.

src/main/java/com/artipie/metrics/MetricsContext.java
dgarus: This is a temporary solution. 
In the result of #1175 we will not have separate parameters host/port for Vert.x metrics.

dgarus:
@olenagerasimova thanks!
I fixed your comments. Please, have a look.

olenagerasimova (approved):
@dgarus thanks

PR #1182, opened 2022-11-14
feat: Getting artipie server port setting via rest api

andpopov:


andpopov:

src/main/resources/swagger-ui/yaml/settings.yaml
```diff
@@ -13,6 +13,25 @@ tags:
   - name: settings
     description: Operations about settings
 paths:
+  /api/v1/settings/port:
+    get:
+      summary: Artipie port.
```
olenagerasimova, line 18: I'd suggest to specify what exactly port is it: `Artipie server-side port (repositories default port)`

olenagerasimova (changes requested):
@andpopov thanks, just minor suggestion. Also, keep in mind, that repository can be served on [individual port](https://github.com/artipie/artipie/wiki/Configuration-Repository#single-repository-on-port)

src/main/resources/swagger-ui/yaml/settings.yaml
andpopov: Thanks, added

olenagerasimova (approved):
@andpopov thanks

Issue #1006, opened 2022-02-11
Measure slice and storage operations


g4s8:
We have `MeasuredSlice` and `MeasuredStorage` to track time for method calls of slice and storage. Now it prints results to logger. Let's change the implementation and submit these metrics to `Metrics` (we'll need to support histogram data type there: https://prometheus.io/docs/concepts/metric_types/#histogram). We are interested in these metrics:
 - [x] execution time of all operations of `Storage` (ms) [histogram]
 - [x] number of each operation in storage [counter]
 - [x] number of chunks in `save` and `load` `Storage` operations (see impl. below) [histogram]
 - [x] number of bytes in `save` and `load` `Storage` operations (see impl. below)
 - [x] number of `Slice` response method calls [counter]
 - [x] number of **request** body chunks per request (see impl. below) [histogram]
 - [x] number of bytes in **request** (see impl. below) [histogram]
 - [x] number of **response** body chunks per request (see impl. below) [histogram]
 - [x] number of bytes in **response** (see impl. below) [histogram]
 - [x] execution time of `Slice.response` method call (ms) [histogram]
 - [x] execution time of `Response.send` method calls (ms) [histogram]
 - [x] execution time of `Connection.accept` method calls (ms) [histogram]

*Implementation notes:*
 - To measure number of chunks in `Publisher` create `Subscriber` decorator with internal counter, increment counter on each `onNext` method call, complete counting on `onComplete`. Create `Publisher` decorator to override `subscribe` method and inject `Subscriber` from 1-st step to origin: `this.origin.subscribe(new MeasuredSubscriber(subscriber))`
 - TO measure number of bytes in `Publisher` create `Subscriber` decorator with internal counter, on each `onNext(ByteBuffer buf)` call: add `buf.remaining()` to counter, on complete publish the results. It could be the same class as from 1-st example.

https://github.com/prometheus/client_java

baudoliver7:
@g4s8 Is it the right repository to work on this task or would it be [artipie/front](https://github.com/artipie/front) or other ?

dgarus:
> @g4s8 Is it the right repository to work on this task or would it be [artipie/front](https://github.com/artipie/front) or other ?

Hi!
[artipie/front](https://github.com/artipie/front) will be deprecated soon.

olenagerasimova:
done

PR #1184, opened 2022-11-23
feat: measure RQ and RS body sizes

olenagerasimova:
Part of #1006 

1) Added all possible [out-of-the-box measures](https://micrometer.io/docs/ref/jvm) micrometer support (jwm: threads, classes, gc etc)
2) added publisher decorator to measure bodies sizes and number of chunks as a separate class (as it will leter be used in storage measures)
3) added slice to measure rq and rs bodies sizes and count requests
4) removed two slices with old metrics

Here is [counter](https://micrometer.io/docs/concepts#_counters) example
`request.counter(COUNTER)[method='GET', route='/same/path', status='OK']; count=2.0`

it has tags with method, route and status.

Here is an example of [summary](https://micrometer.io/docs/concepts#_distribution_summaries) with body sizes:
`artipie.response.body.size(DISTRIBUTION_SUMMARY)[method='GET', route='/same/path']; count=3.0, total=15.0 bytes, max=6.0 bytes`

count here is the number of chunks, total is a total bytes and max is the maximum chunk size. Note, that all micrometer counters with identical names and tags are automatically summarized, so maybe we need to add some identical tag here to distinguish different requests/responses. 

Also, I'm attaching text file with the response from metrics endpoint.


olenagerasimova:
@dgarus please, review

dgarus (approved):
@olenagerasimova thanks!
LGTM

olenagerasimova:
@dgarus please, could you approve again? I've removed failed IT for old `PrometeusSlice`, there is no point in fixing it as it will be removed anyway.

dgarus (approved):

Issue #1183, opened 2022-11-18
Generate JFR events for a storage's operations


dgarus:
Create `Storage` wrapper which will generate JFR events for operations.

PR #1185, opened 2022-11-24
`Content` wrapper that allows to get chunks count and received data size.

dgarus:
part of #1183

- implemented `ChunksAndSizeMetricsContent` allows getting chunks count and received data size.
- tests to check `ChunksAndSizeMetricsContent`


src/test/java/com/artipie/jfr/ChunksAndSizeMetricsContentTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * To check size of content`s data.
+     */
+    private AtomicLong sum;
+
+    /**
+     * To consume data.
+     */
+    private Storage asto;
```
olenagerasimova, line 63: hm, using storage in this test looks a bit of an overhead, I'd suggest using standard methods to process the publisher, for example:
```
Flowable.fromPublisher(content).lastOrError().blockingGet();
```
where content is the `ChunksAndSizeMetricsContent` instance

olenagerasimova (changes requested):
@dgarus please, check my suggestion

olenagerasimova (approved):
@dgarus thanks, looks good

PR #1190, opened 2022-11-28
Generate JFR events for a storage's operations

### LINKED ISSUE #1183: Generate JFR events for a storage's operations

dgarus:
closes #1183 

- created JFR events
- implemented `JfrStorage` wrapper that generates `StrageXXXEvent`s
- created tests


src/main/java/com/artipie/jfr/JfrStorage.java
```diff
@@ -0,0 +1,173 @@
...
+    public JfrStorage(final Storage original) {
+        this.original = original;
+    }
+
+    @Override
+    public CompletableFuture<Boolean> exists(final Key key) {
+        final StorageExistsEvent event = new StorageExistsEvent();
+        event.storage = this.identifier();
+        event.key = key.string();
+        return this.original.exists(key)
```
olenagerasimova, line 43: why we are not counting time here?

src/main/java/com/artipie/jfr/JfrStorage.java
```diff
@@ -0,0 +1,173 @@
...
+                    (chunks, size) -> {
+                        event.chunks = chunks;
+                        event.size = size;
+                        event.commit();
+                    }
+                )
+            );
+        } finally {
+            event.commit();
+        }
```
olenagerasimova, line 72: is it correct to call `commit()` twice? it will first be called in finally-block and when publisher is fully consumed...

src/main/java/com/artipie/jfr/JfrStorage.java
```diff
@@ -0,0 +1,173 @@
...
+        );
+    }
+
+    @Override
+    public CompletableFuture<Void> move(final Key source, final Key target) {
+        final StorageMoveEvent event = new StorageMoveEvent();
+        event.storage = this.identifier();
+        event.key = source.string();
+        event.target = target.string();
+        return this.original.move(source, target)
```
olenagerasimova, line 96: why do not call `begin`?

src/main/java/com/artipie/settings/YamlStorage.java
```diff
@@ -40,6 +41,8 @@ public Storage storage() {
         if (Strings.isNullOrEmpty(type)) {
             throw new IllegalArgumentException("Storage type cannot be null or empty.");
         }
-        return new Storages().newStorage(type, this.yaml);
+        return new JfrStorage(
```
olenagerasimova, line 44: I do not think it's correct to use `JfrStorage` as we should not measure anything if it's not configured

src/main/java/com/artipie/jfr/JfrStorage.java
olenagerasimova: if we want to commit the event in the case of error it will be better to use `handle()` method

olenagerasimova (changes requested):
@dgarus thanks, added comments and here are two more points:
1) why some operations include time count and some not?
2) I'd suggest to wait a bit with this decorator usage, I'm performing some refactoring now to structure all the places where storage is created. Also, we need to use the decorator only if such metric is configured

src/main/java/com/artipie/jfr/JfrStorage.java
dgarus: it's bug

src/main/java/com/artipie/jfr/JfrStorage.java
dgarus: I forget to delete this after debug ((

src/main/java/com/artipie/settings/YamlStorage.java
dgarus: JFR recording can be switched on at any time, so we always have to create wrapper.

dgarus:
@olenagerasimova thanks!
I fixed your comments, please, have a look.

olenagerasimova (dismissed):
@dgarus thanks

andpopov (changes requested):

src/main/java/com/artipie/jfr/JfrStorage.java
```diff
@@ -0,0 +1,173 @@
...
+        event.storage = this.identifier();
+        event.key = key.string();
+        event.begin();
+        try {
+            return this.original.save(
+                key,
+                new ChunksAndSizeMetricsContent(
+                    content,
+                    (chunks, size) -> {
+                        event.chunks = chunks;
```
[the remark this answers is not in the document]
dgarus: > The problem is that may be heppen that second thread will not see changes of first thread.

I don't think there is a problem. 
Each field gets a value only one time, there is no a data race.

olenagerasimova (approved):

andpopov:
@dgarus Thanks!

PR #1196, opened 2022-11-29
feat: use micrometer storage

olenagerasimova:
Part of #1006 

Used micrometer storage inside repository settings according to metrics settings. 

We, actually, have two more places, where storage is used - the settings itself (so-called settings storage) and rest API, but I do not think operations in these storages are should be monitored.

Documentation will be added in the next PR.


dgarus (approved):
@olenagerasimova thanks!

PR #1197, opened 2022-12-01
feat: response timers and docs

olenagerasimova:
Part of #1006

Added timers for response actions (as requested in the ticket) and description of the metrics. Also, attaching metrics response example. 

[promu.txt](https://github.com/artipie/artipie/files/10130334/promu.txt)


dgarus (approved):
@olenagerasimova thanks!
LGTM

Issue #1192, opened 2022-11-28
Generate JFR events for creating `Storage`


dgarus:
Generate JFR events to track creating instances of `Storage`.

PR #1200, opened 2022-12-01
all caches should be gotten through 'StoragesCache'

dgarus:
part of #1192

- `Storages` is used only in `StoragesCache`
- `StoragesCache` is used to obtain any repository storage


src/main/java/com/artipie/settings/StorageAliases.java
```diff
@@ -32,14 +35,38 @@ public interface StorageAliases {
...
+    Storage storage(StoragesCache cache, String alias);
+
+    /**
+     * Finds or create storage by config.
+     *
+     * @param cache Storages cache
+     * @param node Storage config
+     * @return Storage instance
+     */
+    default Storage storage(StoragesCache cache, YamlNode node) {
```
olenagerasimova, line 53: It does not seem like `StorageAliases` actually need to be interface... Maybe we can convert it to class with empty ctor? In this case it seems like we can get rid of the `StorageAliases` field in several classes...

olenagerasimova (changes requested):
@dgarus thanks, please, check my suggestion, what do you think?

olenagerasimova (approved):
@dgarus thanks

PR #1207, opened 2022-12-05
JFR event for http request

### LINKED ISSUE #1202: JFR events for HTTP requests
Create JFR events for requests to track time for method calls of slice.

dgarus:
closes #1202 

- added `SliceResponseEvent` and `JfrSlice`
- implemented `JfrSliceTest` to check `JfrSlice`

### DIFF
--- a/src/main/java/com/artipie/http/BaseSlice.java
+++ b/src/main/java/com/artipie/http/BaseSlice.java
@@ -5,6 +5,7 @@
 package com.artipie.http;
 
 import com.artipie.http.slice.LoggingSlice;
+import com.artipie.jfr.JfrSlice;
 import com.artipie.micrometer.MicrometerSlice;
 import com.artipie.settings.MetricsContext;
 import java.util.logging.Level;
@@ -28,7 +29,11 @@ public final class BaseSlice extends Slice.Wrap {
     public BaseSlice(final MetricsContext mctx, final Slice origin) {
         super(
             BaseSlice.wrapToBaseMetricsSlices(
-                mctx, new SafeSlice(new LoggingSlice(Level.INFO, origin))
+                mctx, new JfrSlice(
+                    new SafeSlice(
+                        new LoggingSlice(Level.INFO, origin)
+                    )
+                )
             )
         );
     }
--- a/src/main/java/com/artipie/jfr/JfrSlice.java
+++ b/src/main/java/com/artipie/jfr/JfrSlice.java
@@ -1,0 +1,154 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/artipie/LICENSE.txt
+ */
+package com.artipie.jfr;
+
+import com.artipie.http.Connection;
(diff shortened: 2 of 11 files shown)


src/main/java/com/artipie/jfr/JfrSlice.java
```diff
@@ -0,0 +1,154 @@
...
+                    event.requestSize = size;
+                }
+            )
+        );
+        return new JfrResponse(
+            res,
+            (chunks, size) -> {
+                event.responseChunks = chunks;
+                event.responseSize = size;
+                event.commit();
```
andpopov, line 68: Please correct me if I see imposible case or not may be it is not important case, any way I describe my point:
Lets image that during some error happens and the callback function on line 65 will not be called.

In this case we will have input traffic not mentioned in Jfr statistics.
We can consider the definition of two JFR events: SliceRequestEvent and SliceResponseEvent:
 1. SliceRequestEvent {
    public volatile String method;
    public volatile String path;
    public volatile int requestChunks;
    public volatile long requestSize;
 }
 2. SliceResponseEvent {
    public volatile String method;
    public volatile String path;
    public volatile int responseChunks;
    public volatile long responseSize;
 }
 In case error in Jfr statistics will be present SliceRequestEvent.
 
PS: I mention that it is just theoretic thoughts that may be not important for this case.

src/main/java/com/artipie/http/BaseSlice.java
```diff
@@ -28,7 +29,11 @@ public final class BaseSlice extends Slice.Wrap {
     public BaseSlice(final MetricsContext mctx, final Slice origin) {
         super(
             BaseSlice.wrapToBaseMetricsSlices(
-                mctx, new SafeSlice(new LoggingSlice(Level.INFO, origin))
+                mctx, new JfrSlice(
+                    new SafeSlice(
```
andpopov, line 33: Should the JfrSlice be configurable to enable/disable?

andpopov (dismissed):

dgarus:
@andpopov thanks for the review!
I fixed event enabled check, please, have a look.

src/main/java/com/artipie/jfr/JfrStorage.java
```diff
@@ -37,136 +45,230 @@ public JfrStorage(final Storage original) {
...
-                res -> {
-                    event.commit();
-                    return res;
-                }
-            );
+        if (event.isEnabled()) {
+            event.begin();
+            res = this.original.exists(key)
+                .thenApply(
+                    rt -> this.eventProcess(rt, key, event, JfrStorage.EMPTY_RUNNABLE)
```
andpopov, line 54: Could we use more sensible name for 'rt' variable, for example 'exists'?

andpopov (changes requested):
@dgarus Thanks a lot!
I have added one little comment

src/main/java/com/artipie/jfr/JfrStorage.java
dgarus: ok, done

PR #1215, opened 2022-12-12
refactor `StorageAliases` interface

### LINKED ISSUE #1214: Rework `StorageAliases` interface
Let's refactor `StorageAliases` interface as
1) it has public static method
2) only one implementation (the second one just throw some exception)
3) default method

and `StorageAliases` seem to do two completely different jobs: 1) find aliases settings for repository 2) build storage from yaml node

thus it's probably will be better to separate this interface into two classes and add proper tests.

olenagerasimova:
closes #1214 

`StorageAliases` interface was divided into two classes, default method was moved into `StoragesCache` (it seems like a more logical fit there), added test.


src/main/java/com/artipie/settings/AliasSettings.java
```diff
@@ -0,0 +1,55 @@
...
+import com.artipie.misc.ContentAsYaml;
+import hu.akarnokd.rxjava2.interop.SingleInterop;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * Find aliases settings for repository.
+ * @since 0.28
+ */
+public final class AliasSettings {
```
dgarus, line 19: This class has only one method.
Looks like we can make this method static and move it to another place.

dgarus (changes requested):
@olenagerasimova thanks!
Have a look at my comment, please.

src/main/java/com/artipie/settings/AliasSettings.java
olenagerasimova: I do not see what is wrong with the class (or interface) with one method. For example, `Slice`, `Response`, `Function` have one method, are there any problems?

This class has its individual purpose (to find aliases settings from storage) and corresponding test. Besides, public static methods are prohibited. But I think it would be better and more logical to make `Storage` field of the class instead of method parameter.

dgarus: > But I think it would be better and more logical to make Storage field of the class instead of method parameter.

ok

PR #1218, opened 2022-12-13
feat: removed routes and keys from micrometer

### LINKED ISSUE #1216: Add functionality to disable metrics
Let's allow users to disable/enable metrics types (jvm, storage, slice). 

Also, let's remove `route` and `key` tags from all the measurements.

olenagerasimova:
closes #1216 

removed route and key tags from micrometer measurements, updated wiki


dgarus (approved):
@olenagerasimova Thanks!

PR #1220, opened 2022-12-16
Fixed #1213 - powermock-api-mockito2 seemed to be superfluous

### LINKED ISSUE #1213: powermock-api-mockito2 seemed to be superfluous
I've checked to remove the dependency `powermock-api-mockito2` and ran the tests locally and it looks the dependency seemed to be not needed anymore. Furthermore the comment of the test class `HttpClientSettingsTest` implies that it is needed which seemed to be proven wrong. Or have I overlooked something?

khmarbaise:
- Removed dependency to powermock-api-mockito2 because it is not used anywhere.
- Removed the todo comment in HttpClientSettingsTest because the describe class does not exist anymore.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -407,12 +407,6 @@ SOFTWARE.
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.powermock</groupId>
-      <artifactId>powermock-api-mockito2</artifactId>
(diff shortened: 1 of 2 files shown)


olenagerasimova (approved):
@khmarbaise thanks

PR #1222, opened 2022-12-19
Support build & publish to maven central

ChGen:
Support build & publish to maven central.

Currently, java versions are set explicitly, or it may give errors due to defaulting to Java 8.

### DIFF
--- a/.github/workflows/maven-central.yml
+++ b/.github/workflows/maven-central.yml
@@ -1,0 +1,63 @@
+name: Create Maven release
+on:
+  push:
+    tags:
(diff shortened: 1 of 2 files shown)


ChGen:

khmarbaise:
The JDK should be used which is currently used to build (JDK11 sure ...why not even JDK17 it is the current LTS)...I would suggest to upload the fat jar to make possible to download the artifact from central and use it in automatic setups for example via Ansible or alike...

olenagerasimova (approved):
@ChGen thanks

Issue #1227, opened 2022-12-23
General token authentication 


olenagerasimova:
Several adapters (at least npm and conda) requires from us to support token authentication. Let's use [vert.x jwt auth](https://vertx.io/docs/vertx-auth-jwt/java/) for these purposes as a) jwt are stateless, b) we do not need to invent any bicycles. Users will be able to use the token in all artipie: as for our rest api and for some repository.

olenagerasimova:
done

PR #1228, opened 2022-12-23
feat: validate JWT tokens

olenagerasimova:
Part of #1227 

Added implementation of `TokenAuthentication` from our `http` to authenticate user by JWT token issued by our rest api. This implementation will be used in npm-adapter later. 

Also, 
- added user groups to the token as they are part of authorization process 
- corrected `com.artipie.api.RestApi#addJwtAuth` method and its usage to use auth in all rest api methods


dgarus (approved):
Thanks!
LGTM

PR #1231, opened 2022-12-28
feat: use token auth in npm-adapter

olenagerasimova:
part of #1227

Updated http and npm, used token auth and added corresponding test.


dgarus (approved):
@olenagerasimova thanks!
LGTM

PR #1232, opened 2022-12-30
feat: generate tokens

olenagerasimova:
Part of #1227 
Implemented `com.artipie.http.auth.Tokens` interface, added test and used it in rest api (so tokens are generated in one place).


dgarus (approved):
@olenagerasimova thanks! lgtm

PR #1235, opened 2023-01-17
doc: typo

Swizbiz:


dgarus (changes requested):
@Swizbiz thanks!
I think, this attribute does not exist. 
Can we delete it from tables?

Swizbiz:
@dgarus it is standard field of events
image

dgarus (approved):
Thanks!
LGTM

Issue #1237, opened 2023-01-26
Integrate permissions model


olenagerasimova:
Let's integrate new permission model from `http.secutiry` package step by step:

- [x] remove unused `Users` interface and implementations #1239 
- [x] cache authenticated users properly #1243 
- [x] implement authentication from storage with new users yaml files structure support #1246 
- [x] update http and files-adapter and integrate new policy  (#1238)
- [x] re-implement resp-api method to update users (#1253)
- [x] implement resp-api method to manage roles (#1255)
- [x] update other adapters (done all)
- [x] implement docker permissions (in docker-adapter https://github.com/artipie/docker-adapter/issues/550)
- [x] prepare detailed policy and permissions documentation (#1257)

PR #1238, opened 2023-01-27
feat: security policy added

olenagerasimova:
part of #1237 

What was done:
- Credentials (users info) format and authentication is changed, now `AuthFromStorage` is used by all adapters. This also means, that `_credentials.yaml` file is not used or required any more
- Extracted all authentication and policy related code into new `ArtipieAuthorization` interface. That's an interface to allow some dummy test implementations. The implementation in the main code parses yaml and instantiates actual auth and policy. Storage will be used in rest api to change user info and permissions
- `SettingsFromPath` class is corrected to add users files on start, configs in `src/resources` are also corrected
- There are two adapters, which are already using new permissions - files and conda
- corrected `TestDeployment` and integration test
- examples (smoke test) configuration corrected

What is not done:
- rest api for now cannot work with new users/roles info structure, will be done in the next PR
- DebianIT with gpg sign is failing. Does not seem to be related with policy changes, I'll fix it during review (done)


src/main/java/com/artipie/adapters/docker/DockerPermissions.java
```diff
@@ -32,7 +32,7 @@ public DockerPermissions(final Permissions origin) {
     }
 
     @Override
-    public boolean allowed(final Authentication.User user, final String action) {
+    public boolean allowed(final AuthUser user, final String action) {
```
dgarus, line 35: I don't like this `DockerPermissions` class.
It looks like bottleneck, and we should get rid of this class later.

src/main/java/com/artipie/auth/YamlPermissions.java
```diff
@@ -50,24 +49,11 @@ public YamlPermissions(final YamlMapping yaml) {
     }
 
     @Override
-    public boolean allowed(final Authentication.User user, final String action) {
+    public boolean allowed(final AuthUser user, final String action) {
```
dgarus, line 52: This class should be removed. Isn't it?

src/main/java/com/artipie/settings/ArtipieAuthorization.java
```diff
@@ -0,0 +1,303 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import org.keycloak.authorization.client.Configuration;
+
+/**
+ * Artipie authorization: authentication and permissions policy.
+ * @since 0.29
+ */
+public interface ArtipieAuthorization {
```
dgarus, line 36: The name `ArtipieAuthorization` doesn't reflect class's purpose

src/main/java/com/artipie/settings/ArtipieAuthorization.java
```diff
@@ -0,0 +1,303 @@
...
+
+    /**
+     * Instance of {@link CachedUsers} which implements
+     * {@link Authentication} and {@link com.artipie.asto.misc.Cleanable}.
+     * @return Cached users
+     */
+    CachedUsers authentication();
```
dgarus, line 43: Why doesn't this method return `Authentication` interface?

dgarus (changes requested):
Thanks!
Please take a look at my comments.

src/main/java/com/artipie/settings/ArtipieAuthorization.java
olenagerasimova: Any suggestions?

src/main/java/com/artipie/auth/YamlPermissions.java
olenagerasimova: It will be removed as soon as all adapter will migrate to new permissions model, including docker (and for docker we need to implement its own permissions).

src/main/java/com/artipie/adapters/docker/DockerPermissions.java
olenagerasimova: We will get rid of it as soon as we implement docker permissions in docker adapter.

src/main/java/com/artipie/api/UsersRest.java
```diff
@@ -73,14 +75,15 @@ public void init(final RouterBuilder rbr) {
      * @param context Request context
      */
     private void deleteUser(final RoutingContext context) {
+        final String uname = context.pathParam(RepositoryName.UNAME);
```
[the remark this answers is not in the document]
olenagerasimova: I've rename the variables, but note that we are using `uname` and `rname` in the swagger [yaml and docs descriptions](https://github.com/artipie/artipie/blob/master/src/main/resources/swagger-ui/yaml/repo-org.yaml#L73) for the path parts for user name and repo name.

src/main/java/com/artipie/auth/AuthFromKeycloak.java
```diff
@@ -47,7 +48,7 @@ public Optional<User> user(final String username, final String password) {
             final Set<String> roles = new HashSet<>();
             roles.addAll(AuthFromKeycloak.realmRoles(token));
             roles.addAll(AuthFromKeycloak.clientRoles(token));
-            return Optional.of(new User(username, roles.stream().toList()));
+            return Optional.of(new AuthUser(username, "keycloak"));
```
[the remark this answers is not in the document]
olenagerasimova: yes, we do not have any mechanism to deal with these roles from keyclock. Custom policy implementation is required to somehow process and deal with them.

src/main/java/com/artipie/settings/ArtipieAuthorization.java
olenagerasimova: because we need both `Authentication` to authenticate and `Cleanable` to reload users cash when user info is updated

src/main/java/com/artipie/auth/LoggingAuth.java
```diff
@@ -45,8 +46,8 @@ public LoggingAuth(final Authentication origin, final Level level) {
     }
 
     @Override
-    public Optional<Authentication.User> user(final String username, final String password) {
-        final Optional<Authentication.User> res = this.origin.user(username, password);
+    public Optional<AuthUser> user(final String username, final String password) {
```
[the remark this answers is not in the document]
olenagerasimova: hm, good point. Maybe we can switch into `char[]` when creating factory for `Authentication` interface? The interface itself is in `http` and after signature is changed we will to update a`artipie/artipie` and several adapters again. What do you say?

andpopov (changes requested):

src/main/java/com/artipie/settings/ArtipieAuthorization.java
dgarus: Can `Authentication` be extended by `Cleanable`?
Or we could create an `CleanableAuthentication` interface.

src/main/java/com/artipie/settings/ArtipieAuthorization.java
dgarus: Maybe `ArtipieSecurity`

dgarus (changes requested):

olenagerasimova:
@dgarus 

> Can Authentication be extended by Cleanable?

I do not thinks it's a good idea as `Authentication` has a lot of implementations where we not need to clean or expire anything.

> Or we could create an CleanableAuthentication interface.

It will have only implementation (current `CachedUsers` class) and you will ask to remove it.

src/main/java/com/artipie/auth/AuthFromKeycloak.java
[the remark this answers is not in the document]
dgarus: In that case, `Set<String> roles` should be removed because it isn't used.

dgarus:
> 
Looks like we have to review interfaces/classes relations.

andpopov:
+ rename 
1. ```Optional<User> Authentication.user(String username, String password)``` to ```Optional<User> Authentication.authenticate(String username, String password)```
2. ```boolean Permissions.allowed(Authentication.User user, String action)``` to ```boolean Permissions.authorize(Authentication.User user, String action)```
or somethig informative?

andpopov:
For example API of Shiro:
```subject.isPermitted("queryPrinter")```

andpopov:
For me it would be convinient to have some static methods to get Authorization and Authentication 
[Here](https://github.com/andpopov/shiro) I have added example with Shiro, Interested sections are 'Authorization' and 'Authentication'

olenagerasimova:
@andpopov 

> boolean Permissions.allowed(Authentication.User user, String action) to boolean Permissions.hasPermissions(Authentication.User user, String action)
or somethig informative?

`Permission` and `Permissions` interfaces and implementations will be removed as soon as all the adapters will migrate into new permissions model

> Optional<User> Authentication.user(String username, String password) to Optional<User> Authentication.login(String username, String password)

The word "login" is a verb and [usually methods are called with the verb returns `void`](https://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#method-names), not some object. 

> For me it would be convinient to have some static methods to get Authorization and Authentication
[Here](https://github.com/andpopov/shiro) I have added example with Shiro, Interesting sections are 'Authorization' and 'Authentication'

Public static methods are not welcomed in artipie and also what for are you going to use these methods? We already have slices to handle authz in the adapters, you can check the example [here](https://github.com/artipie/maven-adapter/pull/263) and [here](https://github.com/artipie/npm-adapter/pull/374).

dgarus (approved):
Thanks!

andpopov (approved):
@olenagerasimova 
Looks good, thanks!

PR #1239, opened 2023-01-27
feat: removed Users interface and implementations

olenagerasimova:
Part of #1237

Removed `Users` interface and implementations as now the only used method was the method to get `Authentication`. Earlier the
interface was used in the old REST API methods.


dgarus (approved):
lgtm

PR #1243, opened 2023-02-13
feat: cache authenticated users

olenagerasimova:
part of #1237

As now credentials yaml is parsed on each authentication request, I've re-implemented the caching:
1. `CachedUsers` now implements `Authentication` interface as each adapter slice accepts `Authentication`. It also implements new `Cleanable` interface to provide opportunity to clean itself (required when user was changed via REST API)
2. `Authentication` instance is now created and initialized only once on start (in the `YamlSettings` ctor) and `Settings#auth` became plain blocking method.
3. `ArtipieCaches` is also now created in the `YamlSettings` ctor, getter for `ArtipieCaches` `Settings#caches` allowed to simplify several other ctors.
4. `CredsConfigCache` and its implementation will be removed as soon as we move to the new users' credential model (when each user has its own separate file with creds and other info https://github.com/artipie/http/issues/501)
5. `ArtipieCaches` now return users cache as `Cleanable` (used in REST API to clean the cache). Later it will also provide opportunity to reload policy caches.
6. `Settings.Fake` was moved to test scope

This approach should lower yaml parsing operations in the case auth from yaml is used and request has `Authorization` header (one parsing per one request). To check it, perform fixed amount of requests on master and on Artipie built from this branch with yaml auth and adding `Authorization` header.


src/main/java/com/artipie/settings/cache/CachedUsers.java
```diff
@@ -47,94 +55,39 @@ final class CachedUsers implements AuthCache {
...
         final String username,
-        final String password,
-        final Authentication origin
+        final String password
     ) {
-        final Data data = new Data(username, password, origin);
+        final byte[] key = DigestUtils.md5(String.join(" ", username, password));
         return new UncheckedScalar<>(
-            () -> this.users.get(data, data::user)
+            () -> this.users.get(key, () -> this.origin.user(username, password))
```
dgarus, line 76: `byte[]` as a key is a bad solution
```
@Test
void aaa(){
    byte[] a = "aaa".getBytes();
    byte[] b = "aaa".getBytes();

    System.out.println(a.equals(b));
    System.out.println(a.hashCode());
    System.out.println(b.hashCode());
}
```
Output:
false
1238959340
76432244

dgarus (changes requested):
Thanks!
Please look at my comment.

olenagerasimova:
@dgarus my bad, used hex string and added test.

dgarus (approved):
lgtm

olenagerasimova:
@Swizbiz will you check it or I can merge?

Swizbiz (approved):
@olenagerasimova thank you. 
Look good to me.

PR #1246, opened 2023-02-17
feat: auth from storage

olenagerasimova:
part of #1237 

Added `Authentication` implementation to check user password from storage. It will be used with new policy/permissions implementation.

Also, I think it would be better to correct `Authentication` and `Policy` interfaces to return `CompletionStage`. Now implementations use `BlockingStorage` and thus blocking operations occurs in the chain of non-blocking requests. What do you say?

### DIFF
--- a/src/main/java/com/artipie/auth/AuthFromStorage.java
+++ b/src/main/java/com/artipie/auth/AuthFromStorage.java
@@ -1,0 +1,106 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/artipie/LICENSE.txt
+ */
+package com.artipie.auth;
+
+import com.amihaiemil.eoyaml.Yaml;
+import com.amihaiemil.eoyaml.YamlMapping;
+import com.artipie.asto.Key;
+import com.artipie.asto.blocking.BlockingStorage;
+import com.artipie.http.auth.Authentication;
+import com.jcabi.log.Logger;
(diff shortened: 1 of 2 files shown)


ChGen (approved):
@olenagerasimova , looks good to me. As for refactoring those interfaces to non-blocking APIs, I support that, since we intend to support wide range of storage types and authentication methods.

src/main/java/com/artipie/auth/AuthFromStorage.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Ctor.
+     * @param asto Abstract blocking storage
+     */
+    public AuthFromStorage(final BlockingStorage asto) {
+        this.asto = asto;
+    }
+
+    @Override
+    public Optional<User> user(final String name, final String pass) {
```
dgarus, line 62: As I understand, `User` class was removed from `http`.
Is this implementation temporary?

olenagerasimova: hm, the class was renamed into `AuthUser` and moved to the upper level, as soon as http is released we will have to correct all the `Authentication` implementations.

olenagerasimova: The corrections will be minimal, just change the class name

dgarus (approved):
Thanks! lgtm

I suggest making PRs complete because it is hard to review.

PR #1251, opened 2023-03-03
Self-hosted GitHub action for benchmarks (WIP)

ChGen:
This is self-hosted GitHub action for benchmarks. It's still WIP, but I have to run it in real CI pipeline. Part of https://github.com/artipie/benchmarks/issues/64


ChGen:
Currently working on this via personal fork of the repository, to reduce possible issues. All remarks are welcomed.

ChGen:
> We need to discuss when exactly we will run the benchmarks? After the release of the fresh artipie docker image? If so, there is no need to set version and build artipie with `mvn version:set` and `mvn install`...

@olenagerasimova , good point. Currently I'd like to support this action for any pushed tag. For this I'd have to make `artipie/artipie:1.0-SNAPSHOT` in the action and run tests for it.

PR #1253, opened 2023-03-10
feat: re-implement users rest api

olenagerasimova:
part of #1237 

Corrected users rest api to comply with new users yamls structure. Also, added methods to enable and disable users.

### DIFF
--- a/src/main/java/com/artipie/api/ManageUsers.java
+++ b/src/main/java/com/artipie/api/ManageUsers.java
@@ -10,27 +10,31 @@ import com.amihaiemil.eoyaml.YamlMappingBuilder;
 import com.amihaiemil.eoyaml.YamlNode;
 import com.artipie.asto.Key;
 import com.artipie.asto.blocking.BlockingStorage;
+import com.artipie.asto.ext.KeyLastPart;
+import com.artipie.asto.misc.UncheckedIOFunc;
 import com.artipie.misc.Json2Yaml;
 import com.artipie.misc.Yaml2Json;
-import com.artipie.settings.YamlSettings;
 import com.artipie.settings.users.CrudUsers;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Collections;
 import java.util.Optional;
-import java.util.stream.Collectors;
 import javax.json.Json;
 import javax.json.JsonArray;
 import javax.json.JsonArrayBuilder;
 import javax.json.JsonObject;
 import javax.json.JsonObjectBuilder;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 
 /**
  * Users from yaml file.
  *
  * @since 0.1
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  */
+@SuppressWarnings("PMD.TooManyMethods")
 public final class ManageUsers implements CrudUsers {
 
     /**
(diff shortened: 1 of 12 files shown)


src/main/java/com/artipie/api/ManageUsers.java
```diff
@@ -56,123 +55,173 @@ public final class ManageUsers implements CrudUsers {
...
+
+    /**
+     * Copy user info from yaml.
+     * @param yaml Yaml to copy info from
+     * @return Builder with existing data
+     */
+    private static YamlMappingBuilder copyUserInfo(final YamlMapping yaml) {
+        YamlMappingBuilder builder = Yaml.createYamlMappingBuilder();
+        for (final YamlNode node : yaml.keys()) {
+            final String val = node.asScalar().value();
```
andpopov, line 175: more sutable name is 'key' or 'uname' instead of 'val'

src/main/java/com/artipie/api/ManageUsers.java
```diff
@@ -56,123 +55,173 @@ public final class ManageUsers implements CrudUsers {
...
-                        new String(this.blsto.value(this.key), StandardCharsets.UTF_8)
-                    ).readYamlMapping().yamlMapping("credentials")
-                );
-            } catch (final IOException err) {
-                throw new UncheckedIOException(err);
-            }
+    private Optional<YamlMapping> userInfo(final String uname) {
+        Optional<byte[]> res = Optional.empty();
+        final Pair<Key, Key> keys = ManageUsers.keys(uname);
+        if (this.blsto.exists(keys.getLeft())) {
```
andpopov, line 155: Can we use: userFileKey ?

src/main/java/com/artipie/api/RestApi.java
```diff
@@ -152,7 +152,12 @@ private void startServices(final RouterBuilder repoRb, final RouterBuilder userR
             new RepoData(this.configsStorage, this.caches.storagesCache()), this.layout
         ).init(repoRb);
         new StorageAliasesRest(this.caches.storagesCache(), asto, this.layout).init(repoRb);
-        Logger.warn(this, "Users API is not available for now");
+        if (this.security.policyStorage().isPresent()) {
```
andpopov, line 155: Why we active user management in REST only if security policyStorage is present?

olenagerasimova: Absence of `policyStorage` means that `artipie` credentials format and policy is not used and we do know the format or place where and how users info and policy is stored and thus cannot manage it.

src/main/java/com/artipie/api/ManageUsers.java
```diff
@@ -56,123 +55,173 @@ public final class ManageUsers implements CrudUsers {
...
-        users.map(
-            yaml -> yaml.keys().stream().map(node -> node.asScalar().value()).map(
-                name -> jsonFromYaml(name, users.get().yamlMapping(name))
-            ).collect(Collectors.toList())
-        ).orElse(Collections.emptyList()).forEach(builder::add);
+        try {
+            for (final Key key : this.blsto.list(new Key.From("users"))) {
+                builder.add(
+                    jsonFromYaml(
+                        new KeyLastPart(key).get().replace(".yaml", "").replace(".yml", ""),
```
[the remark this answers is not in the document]
olenagerasimova: hm, the task to determine and remove the extension is not that simple and obvious (as there are double extensions and hidden folders) and we should not forget that `Key` does not always represent file. So, I think it's better not to generalize and remove what's necessary.

src/main/java/com/artipie/api/ManageUsers.java
olenagerasimova: no, because it does not check that the second key variant actually exists

olenagerasimova:
@andpopov thanks, corrected and answered

dgarus (approved):
Thanks!

andpopov (approved):
@olenagerasimova thanks!

Issue #1262, opened 2023-04-03
Custom authentication


olenagerasimova:
Let's add factories and loader for authentication to provide possibility to implement authentication of user's choice and needs.

olenagerasimova:
done

PR #1263, opened 2023-04-04
feat: add auth factories

olenagerasimova:
part of #1262 

Added factories for authentications and tests.


src/main/java/com/artipie/settings/YamlSettings.java
```diff
@@ -214,9 +168,10 @@ private static CachedUsers initAuth(final YamlMapping settings) {
             );
             res = new AuthFromEnv();
         } else {
+            final AuthLoader loader = new AuthLoader();
             final List<Authentication> auth = creds.values().stream().map(
-                node -> CredentialsType.valueOf(node.asMapping())
-            ).map(type -> type.auth(settings)).toList();
+                node -> node.asMapping().string(YamlSettings.NODE_TYPE)
+            ).map(type -> loader.newObject(type, settings)).toList();
             res = auth.get(0);
```
andpopov, line 175: Can we add check of size of 'auth' list? Because if auth is empty then IndexOutOfBoundsException will be thrown.

andpopov (changes requested):

src/main/java/com/artipie/settings/YamlSettings.java
olenagerasimova: hm, it's not possible as we checked credentials sequence and it's not empty. Then, we perform mapping.

src/main/java/com/artipie/settings/YamlSettings.java
```diff
@@ -214,9 +168,10 @@ private static CachedUsers initAuth(final YamlMapping settings) {
...
             for (final Authentication users : auth.subList(1, auth.size())) {
```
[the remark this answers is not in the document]
olenagerasimova: list size cannot be less then 1 here and if the size = 1, sublist is empty and for-loop is not entered (check the tests, this case is covered)

andpopov (approved):
@olenagerasimova Looks good, thanks!

src/main/java/com/artipie/settings/YamlSettings.java
```diff
@@ -214,106 +168,76 @@ private static CachedUsers initAuth(final YamlMapping settings) {
...
+                        new StoragesLoader().newObject(
+                            asto.get().string(YamlSettings.NODE_TYPE),
+                            new Config.YamlStorageConfig(asto.get())
+                        )
+                    );
+                } else if (policy != null
+                    && YamlSettings.ARTIPIE.equals(policy.string(YamlSettings.NODE_TYPE))
+                    && policy.yamlMapping(YamlSettings.NODE_STORAGE) != null) {
+                    res = Optional.of(
+                        new StoragesLoader().newObject(
```
dgarus, line 229: `StoragesLoader` problem. Storage factories already initiated for this moment.

olenagerasimova: will be fixed in #1270

dgarus (approved):
Thanks!

PR #1265, opened 2023-04-05
fix: correct default settings copy

olenagerasimova:
part of #1237 

corrected copying of default settings:
1) storage aliases file is located inside repo dir now
2) permissions sections removed from repo settings
3) added read permissions for any repo


andpopov (approved):
@olenagerasimova Thanks!

dgarus (approved):
Thanks!

PR #1269, opened 2023-04-05
fix: correct example config files

olenagerasimova:
part of #1237 

fixed example configs format


olenagerasimova:
@dgarus sorry, forgot to remove upper-level mapping with user/group name

dgarus (approved):
Thanks!

PR #1270, opened 2023-04-05
fix: single storages loader

olenagerasimova:
part of #1237 

Create storages loader instance only once in artipie/artipie

### DIFF
--- a/src/main/java/com/artipie/http/DockerRoutingSlice.java
+++ b/src/main/java/com/artipie/http/DockerRoutingSlice.java
@@ -5,12 +5,18 @@
 package com.artipie.http;
 
 import com.artipie.docker.http.BaseEntity;
+import com.artipie.docker.perms.DockerActions;
+import com.artipie.docker.perms.DockerRepositoryPermission;
 import com.artipie.http.auth.Authentication;
-import com.artipie.http.auth.BasicAuthSlice;
+import com.artipie.http.auth.BasicAuthzSlice;
+import com.artipie.http.auth.OperationControl;
 import com.artipie.http.rq.RequestLine;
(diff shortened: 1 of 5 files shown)


olenagerasimova:
@dgarus please, review, just a view changes

dgarus (dismissed):
thanks!

src/main/java/com/artipie/http/DockerRoutingSlice.java
```diff
@@ -63,10 +71,20 @@ public Response response(final String line, final Iterable<Map.Entry<String, Str
...
             if (group.isEmpty() || group.equals("/")) {
-                rsp = new BasicAuthSlice(
+                rsp = new BasicAuthzSlice(
                     new BaseEntity(),
                     this.settings.authz().authentication(),
-                    user -> !user.equals(Authentication.ANY_USER)
+                    new OperationControl(
+                        user -> {
+                            // @checkstyle NestedIfDepthCheck (10 lines)
+                            PermissionCollection res = new FreePermissions();
```
dgarus, line 80: Can't understand what does this mean?

olenagerasimova: the same as was `user -> !user.equals(Authentication.ANY_USER)`:
`v2` endpoint (docker root) is available for any authenticated user

dgarus: > the same as was `user -> !user.equals(Authentication.ANY_USER)`: `v2` endpoint (docker root) is available for any authenticated user

This means if the user is authenticated, he has the permission to pull any image. Right?

PR #1274, opened 2023-04-11
fix: copy example resources into different storages

olenagerasimova:
part of #1237 

Fixed resources copy process in `SettingsFromPath`: as we now have different storages for repo settings and security, it's necessary to copy resources correspondingly.


dgarus (approved):
Thanks!

Issue #1273, opened 2023-04-06
Rest API permissions


olenagerasimova:
Let's implement permissions for rest api. 
Note, the API allows CRUD operations for repositories, storage aliases, users and roles. There will be no differences for 
api permissions for org and flat layouts, permission to create or update repositories allows user to create or update any repositories in Artipie.
Here is the suggestion:
```yaml
api_repository_permissions:
  - read: # read includes list, view repo settings, checking for existence
  - create
  - move
  - delete
api_storage_aliase_permissions: # storage aliases for some repositories, users or common
  - read
  - create
  - delete
api_user_permissions: # modify users
  - read
  - create
  - update
  - enable # allows enable and disable users
  - change_password
api_role_permissions: # modify roles
  - read
  - create
  - update
  - enable # allows enable and disable
``` 
We also support * (any action is allowed) in for operation list.
Also, if user has `all_permission`, any action in API is automatically allowed.

olenagerasimova:
@dgarus please, check. Here are suggestions for rest api permissions

dgarus:
@andpopov @ChGen 
please, check.

dgarus:
For `api_user_permissions` suggest changes:
 - join `enable/disable` to one action (the same for `api_role_permissions`);
 - add action `change_password`. 
 
User should have the opportunity to change their password.
`api_settings_permissions` looks as excessive permissions for me.
WDYT?

olenagerasimova:
oh, yes, we have `change_password` operation. As for `api_settings_permissions` - yees, let's remove it, every authenticated user will be able to get this info

olenagerasimova:
done

PR #1283, opened 2023-04-26
feat: test for rest api permissions

olenagerasimova:
Part of #1273 

Added test for rest api with permissions and corrected rest api permission collection. As we use heterogeneous collection `java.security.Permissions` each permission class must have his own collection class. 

Also initial collection implementation was not correct because permission key contained action mask. Mask can differ depending on allowed actions list (it can be `create and delete` or `update and enable` or etc), thus checking for action `create` we will never get suitable permission from collection map by key.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,7 @@ SOFTWARE.
     <dependency>
       <groupId>com.artipie</groupId>
       <artifactId>http</artifactId>
-      <version>v1.2.8</version>
+      <version>v1.2.9</version>
     </dependency>
     <dependency>
       <groupId>com.artipie</groupId>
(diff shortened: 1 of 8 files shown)


src/test/java/com/artipie/api/AuthRestOrgTest.java
```diff
@@ -217,12 +260,22 @@ public Authentication authentication() {
...
             @Override
             public Policy<?> policy() {
-                return Policy.FREE;
+                final BlockingStorage asto = new BlockingStorage(AuthRestOrgTest.super.ssto);
+                asto.save(
+                    new Key.From(String.format("users/%s.yaml", AuthRestOrgTest.NAME)),
+                    String.join(
+                        "\n",
+                        "permissions:",
+                        "  adapter_all_permission: {}"
```
dgarus, line 269: We need to rename `adapter_all_permission` to `all_permissions`

dgarus (approved):
thanks!

src/test/java/com/artipie/api/AuthRestOrgTest.java
olenagerasimova: ok, first, we need to correct `http`

PR #1290, opened 2023-05-12
feat: default roles

### LINKED ISSUE #1289: Default roles
As we discussed I'm suggesting to implement default roles depending on auth type. Here is suggested files structure for `artipie` policy type:
```
├── roles
│   ├── default
│   │   ├──  keycloack.yaml
│   │   ├──  env.yml
│   │   ├──  artipie.yaml
│   ├── java-dev.yaml
│   ├── admin.yaml
│   ├── ...
├── users
│   ├── david.yaml
│   ├── jane.yaml
│   ├── ...
```
Subfolder `default` in `roles` folder will contain yamls with default roles permissions for auth type. The name of the file must match auth type, in java code, we store this auth type in `context` field of `AuthUser` object and thus, we know what default role permissions to check. 
Format of this default roles yaml is the same as the format of yaml with usual role.

olenagerasimova:
closes #1289 

Updated http, added description for default roles in wiki and extended example config.

### DIFF
--- a/.wiki/Configuration-Policy.md
+++ b/.wiki/Configuration-Policy.md
@@ -31,6 +31,10 @@ eviction time can be configured with the help of `eviction_millis` field.
 Policy storage is supposed to have the following format:
 ```
 ├── roles
+│   ├── default
+│   │   ├──  keycloack.yaml
+│   │   ├──  env.yml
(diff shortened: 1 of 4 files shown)


dgarus (approved):
Thanks!

PR #1296, opened 2023-06-02
feat: composer instance to use permissions policy

icukeng:
Composer repository was free of permission check. I try to fix it.
There was a simple class constructor used, which set unrestricted access to repo.

I'm not a java developer, so made by analogy with other policy's changes.
And FYI same situation might be with gem repo.

Part of #1237


olenagerasimova (approved):
@icukeng thanks!

PR #1304, opened 2023-07-06
Scripting support

ChGen:
Here's resumed work on scripting support and scheduled scripts. Initially was started by Andrey Popov. This branch is rebased on current master and basic support added for accessing some artipie objects from the scripts. New tests added to check & demonstrate this objects access. 
Since it's a new feature, there's some points to discuss:
1. New classes design. Should we try make it more EO-like style (for codebase consistency) or keep this implementation (with bunch of `@SuppressWarning()`-s.
2. Passing artipie objects to the scripts. Currently, simple implementation is provided with two new tests: `ArtipieSchedulerTest.runCronJobWithSettingsObject()` and `ArtipieSchedulerTest.runCronJobWithReposObject()`.
3. Providing new specific functionality to the scripts for convenience. For example, access to Artifacts DB. Or extra API from Docker adapter to examine and cleanup storage.
4. Providing extra events to scripts. For example, extra Storage decorator could be created to report storage events to scripts.


ChGen:
@olenagerasimova , please review/comment.

src/main/java/com/artipie/scripting/Script.java
```diff
@@ -0,0 +1,224 @@
...
+     * Create precompiled instance of {@link Script}.
+     * @param name Name of scripting engine.
+     * @param script Script code
+     * @return Script
+     */
+    @SuppressWarnings("PMD.ProhibitPublicStaticMethods")
+    static Script newCompiledScript(final String name, final String script) {
+        final ScriptEngine engine = Script.MANAGER.getEngineByName(name);
+        return new PrecompiledScript(engine, script);
+    }
```
olenagerasimova, line 67: Why do we need these static methods? I'd simply added secondary constructors, which accepts string as engine name and call `Script.MANAGER.getEngineByName(name)` in the call to primary ctor.

src/main/java/com/artipie/scripting/PythonScript.java
```diff
@@ -0,0 +1,37 @@
...
+ * https://github.com/artipie/artipie/LICENSE.txt
+ */
+package com.artipie.scripting;
+
+/**
+ * Python Script.
+ *
+ * @since 0.30
+ */
+public interface PythonScript {
```
olenagerasimova, line 12: do not clear, why scripts for languages are interfaces... It's would be more logical to create parent class with two not static methods and with language name field which would be overridden in each accessor. Each accessor will implement only ctor, passing corresponding language name.

src/main/java/com/artipie/scripting/ScriptRunner.java
```diff
@@ -0,0 +1,103 @@
...
+ * Job for running script in quartz
+ *
+ * @since 0.30
+ */
+public final class ScriptRunner implements Job {
+    @Override
+    public void execute(final JobExecutionContext context) throws JobExecutionException {
+        final Settings settings = (Settings) context.getJobDetail().getJobDataMap().get("settings");
+        final Key key = new Key.From(context.getJobDetail().getJobDataMap().getString("key"));
+        final BlockingStorage storage = new BlockingStorage(settings.configStorage());
```
olenagerasimova, line 31: all these parameters from job data map can potentially be nulls. If any of them is null, it would be better to stop the job. Also, it's possible to make them fields and add setters, then quartz will set the values, check https://github.com/artipie/http/pull/551/files#diff-cede6ef368fefbd7e0f78b524ca0ea8109fb5a738824c4f2762c80a4e4f9f707 as an example

src/main/java/com/artipie/scripting/ScriptRunner.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        return res;
+    }
+
+    /**
+     * Obtain extension of filename.
+     * @param filename Name of file.
+     * @return Extension.
+     */
+    private static Optional<String> extension(final String filename) {
```
olenagerasimova, line 93: is not there some standard corresponding method in `java.nio` or `commons`?

src/main/java/com/artipie/scripting/ScriptRunner.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    /**
+     * Create instance of Script by script-file extension and script code.
+     * @param ext Extension of script-file.
+     * @param script Script code.
+     * @return Script instance
+     */
+    private static Optional<Script> script(final String ext, final String script) {
```
olenagerasimova, line 66: also, we could created enum with languages types instead of this switch. Enum can have abstract method to return instance of parent class and each enum item will implement this method by returning corresponding instance of the script class

olenagerasimova (changes requested):
thanks, please, check my suggestions

src/main/java/com/artipie/scheduler/ArtipieScheduler.java
```diff
@@ -0,0 +1,169 @@
...
+import org.quartz.SchedulerException;
+import org.quartz.Trigger;
+import org.quartz.TriggerBuilder;
+import org.quartz.impl.StdSchedulerFactory;
+
+/**
+ * Scheduler.
+ * @since 0.30
+ */
+public final class ArtipieScheduler {
```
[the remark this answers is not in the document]
ChGen: @olenagerasimova , then I'll switch in this branch to `QuartsService`, since it's already merged and improving.

ChGen:
@olenagerasimova , thank you for the comments. I'll switch in this branch to `QuartsService` instead of new scheduler, since `QuartsService` already merged and improving.

olenagerasimova:
@ChGen hm, may be, the better approach will be to move `QuartsService` to artipie from http and extend it with methods, required for scripting. The idea for  `QuartsService` is to hide as much quartz-related (creating jobs and triggers) work as possible. 

But `QuartzJob` abstract class should be left in http, as it it used in the adapters (in npm only for now, but it will be more adapters using it).

ChGen:
@olenagerasimova , sounds reasonable. Since you are working on this part now, should I wait for you to move Quarts classes to artipie from http ?

olenagerasimova:
@ChGen I think you can simply copy `QuartsService` and test from https://github.com/artipie/http/pull/551 and extend it as you wish. Http with `EventProcessor` is already released, so it should be ok. I'll remove `QuartsService` from http in the PR

ChGen:
The test below will fail on `#3` due to `java.io.IOException: No value for key: my-repo.yml`, unless whole line at `#1` commented out. I think it's some kind of bug and `CompletableFuture` from `#1` somehow gets used, unless I'm missing something...
```
  @Test
  void runReposObjectFail2() throws Exception {
      final YamlSettings settings = runCronScript(
              String.join(
                      "\n",
                      "File file = new File('%1$s')",
                      "file.write 'Hello world'"
              )
      );
      final Key result = new Key.From(ArtipieSchedulerTest.RESULTS_PATH);

      // #1
      final Object cfg0 = new RepositoriesFromStorage(settings).config("my-repo"); //.toCompletableFuture().join();

      // #2
      new RepoConfigYaml("maven")
              .withPath("/artipie/test/maven")
              .withPort(7777)
              .withUrl("http://test.url/artipie")
              .saveTo(new FileStorage(this.temp), "my-repo");

      // #3
      final RepoConfig cfg = new RepositoriesFromStorage(settings).config("my-repo").toCompletableFuture().join();
      MatcherAssert.assertThat(
              new String(this.data.value(result)),
              new IsNot<>(new IsEqual<>(cfg.toString())) // #4 - two different files
      );
  }
```

olenagerasimova:
@ChGen is this part of code is from some existing test? As far as I understand, repo config `my-repo` is actually created on blocking call #2. Call #1 (without join) is not blocking, so config `my-repo` can actually be read after blocking call #2 and thus no exception will be thrown. But if you call `join` in #1 then the exception is inevitable.

olenagerasimova:
@ChGen hm, I see now. Can you commit this test in some branch?

ChGen:
@olenagerasimova , yes I will prepare the branch and let you know.

ChGen:
@olenagerasimova , I pushed branch with the failing test. https://github.com/artipie/artipie/tree/ech-failed-test

ChGen:
@olenagerasimova , thanks, I suspected it's some caching, but somehow missed that and thought it was deeper. I think it worth clarifying in `RepositoriesFromStorage` (maybe), since `Repositories` interface looks benign in this matter. What do you think?

src/main/java/com/artipie/VertxMain.java
```diff
@@ -135,6 +144,7 @@ public void stop() {
         for (final VertxSliceServer server : this.servers) {
             server.stop();
         }
+        this.scheduler.stop();
```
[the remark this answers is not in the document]
ChGen: `scheduler.stop()` also required in tests. Should I remove whole `VertxMain.stop()` ?

ChGen:
@olenagerasimova , please review.

src/main/java/com/artipie/scripting/Script.java
```diff
@@ -0,0 +1,251 @@
...
+    class Result {
+        /**
+         * Script context.
+         */
+        private final ScriptContext context;
+
+        /**
+         * Resulting value.
+         */
+        private Object val;
```
olenagerasimova, line 200: Is this field used anywhere? The setter is private and not called inside this class

src/main/java/com/artipie/scripting/ScriptRunner.java
```diff
@@ -0,0 +1,120 @@
...
+        final Map<String, Script.ScriptType> map = Map.of(
```
olenagerasimova, line 91: I'd suggest use loop via enum values instead of map to get script type. Thus the code will work just fine and no changes will be required if we add another script type.

olenagerasimova (changes requested):
@ChGen thanks, couple minor comments

src/main/java/com/artipie/scripting/Script.java
ChGen: I updated comments for the object - it is required to get result value from `.eval()`. 
For `VertxMain.stop()` - should I remove whole method? I don't see its usage too.

olenagerasimova: I think yes, it would be better to remove it, so it do not cause any misunderstanding

ChGen:

olenagerasimova (approved):
@ChGen thanks, also, I think we should Denis for review

ChGen:
@olenagerasimova , thanks, I added Denis for review too.

ChGen:
@dgarus , please review.

src/test/java/com/artipie/scripting/GroovyScriptTest.java
```diff
@@ -0,0 +1,56 @@
...
+ * Tests for Groovy support by {@link Script}.
+ *
+ * @since 0.30
+ * @checkstyle MagicNumberCheck (500 lines)
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+public class GroovyScriptTest {
+    @Test
+    public void standardScript() throws ScriptException {
+        new StandardScript(ScriptType.GROOVY, "println(1)").call();
```
dgarus, line 27: What does this line check?

src/test/java/com/artipie/scripting/GroovyScriptTest.java
```diff
@@ -0,0 +1,56 @@
...
+        new StandardScript(ScriptType.GROOVY, "println(a)").call(Map.of("a", 2));
```
dgarus, line 28: What does this line check?

src/test/java/com/artipie/scripting/GroovyScriptTest.java
```diff
@@ -0,0 +1,56 @@
...
+        final Map<String, Object> variables = new HashMap<>();
+        variables.put("a", 4);
+        MatcherAssert.assertThat(
+            new StandardScript(ScriptType.GROOVY, "a = a * 3").call(variables).variable("a"),
+            new IsEqual<>(12)
+        );
+    }
+
+    @Test
+    public void precompiledScript() throws ScriptException {
```
dgarus, line 42: This test is the same as above, we could make a method with `Script` type parameter and avoid code duplication.

src/main/java/com/artipie/scheduling/ScriptScheduler.java
```diff
@@ -0,0 +1,134 @@
...
+                                    data.put("settings", settings);
+                                    try {
+                                        this.service.schedulePeriodicJob(
+                                            cronexp, ScriptRunner.class, data
+                                        );
+                                    } catch (final SchedulerException ex) {
+                                        throw new ArtipieException(ex);
+                                    }
+                                }
+                                return null;
```
dgarus, line 129: imho, this id a bad solution. Think, we should filter the stream values.

dgarus (changes requested):
Thanks, please check my comments

src/main/java/com/artipie/scripting/ScriptRunner.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @since 0.30
+ */
+public final class ScriptRunner implements Job {
+    @Override
+    public void execute(final JobExecutionContext context) throws JobExecutionException {
+        final Settings settings = (Settings) context.getJobDetail().getJobDataMap().get("settings");
+        final Key key = new Key.From(context.getJobDetail().getJobDataMap().getString("key"));
+        final BlockingStorage storage = new BlockingStorage(settings.configStorage());
+        if (settings == null || key == null || storage == null) {
```
dgarus, line 37: Condition 'settings == null || key == null || storage == null' is always 'false'

src/main/java/com/artipie/scripting/ScriptRunner.java
```diff
@@ -0,0 +1,116 @@
...
+ * Script runner.
+ * Job for running script in quartz
+ *
+ * @since 0.30
+ */
+public final class ScriptRunner implements Job {
+    @Override
+    public void execute(final JobExecutionContext context) throws JobExecutionException {
+        final Settings settings = (Settings) context.getJobDetail().getJobDataMap().get("settings");
+        final Key key = new Key.From(context.getJobDetail().getJobDataMap().getString("key"));
```
dgarus, line 35: if `context.getJobDetail().getJobDataMap().getString("key")` returns `null` we'll get NPE

ChGen:
@dgarus , @olenagerasimova , currently there are `Script.StandardScript` and `Script.PrecompiledScript` classes for scripting with corresponding tests. Probably its better to keep only one variant, and remove the other as redundant. For example, I can keep `Script.PrecompiledScript` and add some caching implementation of precompiled script objects to the `ScriptRunner` class. Caching could be based on google's `LoadingCache`.

dgarus:
> I can keep Script.PrecompiledScript and add some caching implementation of precompiled script objects to the ScriptRunner class

I agree.
Thanks!

src/main/java/com/artipie/scheduling/ScriptScheduler.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    /**
+     * Loads crontab from settings.
+     * Format is:
+     * <pre>
+     *     meta:
+     *       crontab:
+     *         - key: scripts/script1.groovy
```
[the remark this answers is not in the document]
ChGen: `key` is used as common term for our `Storage`, since some storages don't have file path concept.

src/main/java/com/artipie/scheduling/ScriptScheduler.java
ChGen: Do you mean I should switch to something like `forEach()` instead of `map()`?

ChGen:
@dgarus , please review.

src/main/java/com/artipie/scheduling/ScriptScheduler.java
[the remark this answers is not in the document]
dgarus: > @dgarus , `key` is used as common term for our `Storage`, since some storages don't have file path concept.

Is this a path from a user standpoint? 
Could user know about storage and his keys?

src/main/java/com/artipie/scheduling/ScriptScheduler.java
dgarus: Yes, we don't need to collect nulls.

src/main/java/com/artipie/scripting/ScriptRunner.java
```diff
@@ -0,0 +1,244 @@
...
+    private static final class FilesContent {
+        /**
+         * Key.
+         */
+        private final Key key;
+
+        /**
+         * Storage.
+         */
+        private final BlockingStorage storage;
```
dgarus, line 164: Why do you include `storage` in the key!? 
This doesn't seem right, don't do this.

dgarus (changes requested):
Hi, please check my comments

src/main/java/com/artipie/scripting/ScriptRunner.java
ChGen: @dgarus , currently `key` strongly depends on correct `storage`, since storage objects could be the set of storage wrappers, etc. Here `ScriptRunner` wraps `configStorage` with `BlockingStorage` from the `settings` passed in `JobDataMap`. 
For these reasons I need either holding `storage` instance in addition to the key, or read whole script and use script source code in the `String` form as the key for precompiled scripts cache.

src/test/java/com/artipie/scripting/ScriptingTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private void testScript(
+        final Script.ScriptType type, final String code
+    ) throws ScriptException {
+        final Script script = new Script.PrecompiledScript(type, code);
+        final Map<String, Object> variables = new HashMap<>();
+        variables.put("a", 4);
+        final Script.Result res = script.call(variables);
+        MatcherAssert.assertThat(this.toLong(res.variable("a")), new IsEqual<>(12L));
```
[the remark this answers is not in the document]
ChGen: @dgarus, here I check both ways of accessing variables: with `res.variable()` call, and by using `variables` map passed explicitly in `script.call()`. I check that both ways provide the same results.

src/test/java/com/artipie/scripting/ScriptingTest.java
```diff
@@ -0,0 +1,96 @@
...
+ * Tests Script.StandardScript and Script.PrecompiledScript.
+ * @since 0.1
+ * @checkstyle MagicNumberCheck (200 lines)
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+public class ScriptingTest {
+
+    @Test
+    public void groovyStdTest() throws ScriptException {
+        final String code = "a = a * 3\nb = a + 1";
```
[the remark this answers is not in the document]
ChGen: Ok

src/test/java/com/artipie/scripting/ScriptingTest.java
```diff
@@ -0,0 +1,96 @@
...
+        this.testScript(Script.ScriptType.GROOVY, code);
+    }
+
+    @Test
+    public void groovyPrecompTest() throws ScriptException {
```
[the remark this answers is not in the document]
ChGen: Agree, should be removed.

src/main/java/com/artipie/scripting/ScriptRunner.java
dgarus: > @dgarus , currently `key` strongly depends on correct `storage`, since storage objects could be the set of storage wrappers, etc. Here `ScriptRunner` wraps `configStorage` with `BlockingStorage` from the `settings` passed in `JobDataMap`. For these reasons I need either holding `storage` instance in addition to the key, or read whole script and use script source code in the `String` form as the key for precompiled scripts cache.

Storage cannot be a part of key. This is wrong!

src/test/java/com/artipie/scripting/ScriptingTest.java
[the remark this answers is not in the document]
dgarus: ok

src/main/java/com/artipie/scripting/ScriptRunner.java
ChGen: @dgarus , storage will be removed from `equals()/hashCode()`, but kept inside key for data access.

ChGen:
@dgarus , please review. I also added extra tests for `ScriptRunner.FilesContent`.

ChGen:
@dgarus , @olenagerasimova , actually I'd like to refactor some parts for the branch to get rid of `static` cache and some redundant objects creation.

ChGen:
@olenagerasimova , @dgarus , please review. Here I've got rid of `static` cache of precompiled script instances. `ScriptContext` class now holds cache and required parameters for scripts.

ChGen:
For some reason, integration tests below always fail in CI, but pass locally:
```
Error:  Failures: 
Error:  com.artipie.conda.CondaITCase.canInstallFromArtipie(String,String,String)
Error:    Run 1: CondaITCase.canInstallFromArtipie:85 Package snappy-1.1.3-0 was not installed successfully
Error:    Run 2: CondaITCase.canInstallFromArtipie:85 Package snappy-1.1.3-0 was not installed successfully
Error:  com.artipie.conda.CondaITCase.canUploadToArtipie(String,String,String)
Error:    Run 1: CondaITCase.canUploadToArtipie:115 Failed to set anaconda upload flag
Error:    Run 2: CondaITCase.canUploadToArtipie:115 Failed to set anaconda upload flag
```

ChGen:
Currently debugging this test locally.

ChGen:
Integration tests are passed.

dgarus:
@olenagerasimova 
Could you please review?

src/main/java/com/artipie/scripting/Script.java
```diff
@@ -0,0 +1,226 @@
...
+         */
+        private ScriptContext context() {
+            return this.context;
+        }
+
+        /**
+         * Setter for resulting value.
+         * @param value Resulting value
+         */
+        private void value(final Object value) {
```
olenagerasimova, line 222: I still do not get why the `val` field, **private** setter and **public** getter are required... Setter is not called from anywhere in this class...

src/main/java/com/artipie/scripting/ScriptContext.java
```diff
@@ -0,0 +1,191 @@
...
+         */
+        private static String extension(final String filename) {
+            final int pos = filename.lastIndexOf('.');
+            final String res;
+            if (pos >= 0) {
+                res = filename.substring(pos + 1);
+            } else {
+                res = "";
+            }
+            return res;
```
olenagerasimova, line 188: Cannot we use some standard method to obtain the extension? `FilenameUtils.getExtension` from commons.io should be available

olenagerasimova (changes requested):
@ChGen thanks, just minor suggestions to avoid bicycles and code duplications

src/main/java/com/artipie/scripting/Script.java
ChGen: @olenagerasimova , That setter is called from `com.artipie.scripting.Script.PrecompiledScript#call()`. I renamed it to `setValue()` and added basic test for `eval()` on expression, which also demonstrates usage.

ChGen:
@olenagerasimova , please review.

src/main/java/com/artipie/scripting/ScriptRunner.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    /**
+     * Returns cached script. Scripts objects are handled by LoadingCache.
+     * @param key Key in the `scontext` storage for the script.
+     * @param scontext Target scontext object.
+     * @return Returns cached script or empty Optional in case of error.
+     */
+    private static Optional<Script> cachedScript(final Key key, final ScriptContext scontext) {
```
dgarus, line 88: Caching looks unnecessary complex for me, could you please make it as simple as possible?

dgarus (changes requested):
Thanks. Read please my comment

src/main/java/com/artipie/scripting/ScriptRunner.java
dgarus: We should have something like this
```
scontext.getScripts().get(key, lk -> {
            final String script = new String(scontext.getStorage().value(lk));
            return new Script.PrecompiledScript(script);
        });
```

src/main/java/com/artipie/scripting/ScriptContext.java
```diff
@@ -0,0 +1,176 @@
...
+                    }
+                }
+            );
+    }
+
+    /**
+     * Extra class for caching precompiled scripts.
+     * @since 0.1
+     */
+    static final class FilesContent {
```
dgarus, line 120: I really think we don't need this class.

src/main/java/com/artipie/scripting/ScriptContext.java
```diff
@@ -0,0 +1,176 @@
...
+        /**
+         * Returns precompiled script for stored key & storage.
+         * @return Returns precompiled script object instance.
+         */
+        public Script.PrecompiledScript precompiledScript() {
+            final String ext = FilenameUtils.getExtension(this.key.string());
+            final Script.ScriptType type = Arrays.stream(Script.ScriptType.values())
+                .filter(val -> val.ext().equals(ext)).findFirst().orElse(Script.ScriptType.NONE);
+            Script.PrecompiledScript result = null;
+            if (!type.equals(Script.ScriptType.NONE)) {
```
dgarus, line 169: What is the sense in this check?

src/main/java/com/artipie/scripting/ScriptContext.java
```diff
@@ -0,0 +1,176 @@
...
+                final String script = new String(this.storage.value(this.key));
+                result = new Script.PrecompiledScript(type, script);
```
dgarus, line 171: `Script.PrecompiledScript` should be able to realize what script's type is and returns the result or throws an exception.

src/main/java/com/artipie/scripting/ScriptRunner.java
ChGen: @dgarus , thanks, I agree - now that cache is local, we can rely on constant storage instance.

ChGen:
@dgarus , please review

dgarus (approved):
Thanks! lgtm

We should write documentation.

ChGen:
@dgarus , thanks, I'll provide documentation in the next PR.

andpopov:
👌

PR #1311, opened 2023-09-12
Provide fully functional logger out-of-the-box.

ChGen:
Consider having fully-functional logger out-of-the-box. Since we are small & user-friendly solution, I think we'd better have full logging available in our builds.


ChGen:
@olenagerasimova , please review.

olenagerasimova (approved):
@ChGen thanks

PR #1322, opened 2023-10-26
feat: move adapters into artipie

olenagerasimova:
Moved adapters into this repository and made maven multi module project. Artipie-main module is not an adapter, it's this initial repo. 
All the modules have root-pom as a parent, root-pom contains most of the common dependencies, including `asto` and `http`.
Artipie-main depends on all the adapters in `compile` scope. 
Updated and added scripts for releases: 
- when tag starting with `v` is pushed, artipie docker image is released
- when tag has format `[adapter-name]` **_** `[version]` is pushed, adapter is released in maven-central. 
Example: `conda-adapter_v1.5.0`


ChGen:
@olenagerasimova , I remember we had support for building our adapters for external use with JDK8. Do we still have this feature in this branch ?

dgarus:
@ChGen @olenagerasimova 
I suggest getting rid of `pdd`. We don't use it, but have to spend time fixing errors like this one:
```
ERROR: artipie-main/src/main/resources/swagger-ui/swagger-ui-bundle.js; PDD::Error at artipie-main/src/main/resources/swagger-ui/swagger-ui-bundle.js:2: TODO must have a leading space to become a puzzle, as this page explains: https://github.com/cqfn/pdd#how-to-format
```

olenagerasimova:
Removed `.pdd` scripts and action. 
As for java 8, I've overridden `maven.compiler.target` and `maven.compiler.source` properties in the adapters, which should be build with java 8.

dgarus (approved):
Thanks!

ChGen:
@dgarus , sorry, I pressed wrong Merge button. :( Please revert if necessary.

PR #1326, opened 2023-11-07
feat: move http-client

olenagerasimova:
Moved http-client into artipie and updated jetty version in the client. Also, got rid of [jetty-reactive-httpclient](https://github.com/jetty-project/jetty-reactive-httpclient) usage (as it is too buggy, [one](https://github.com/jetty-project/jetty-reactive-httpclient/issues/322) and [two](https://github.com/jetty-project/jetty-reactive-httpclient/issues/323) ) and switched to async API of jetty 12.0.x (check `JettyClientSlice`).

Also, cleaned poms from developers info. Links to github are left if the common pom.


olenagerasimova:
@dgarus I'm a little stuck here with [jetty-reactive-httpclient](https://github.com/jetty-project/jetty-reactive-httpclient) latest version (it has [several](https://github.com/jetty-project/jetty-reactive-httpclient/issues) problems), what do you think if we get rid of it and use Jetty itself, not its reactive version?

dgarus (approved):
Thanks!

PR #1339, opened 2023-11-23
fix: use --release option

olenagerasimova:
Corrected approach with different java versions for different modules: 
- it's more correct to use [`--release`](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html) javac option instead of `target` and `source` and there is no need to compile with lower java version, [java will do all the magic itself](https://www.baeldung.com/java-compiler-release-option) 
- used `maven-compiler-plugin` to set different release version for tests. Thus, our test code does not have to be compatible with java 8, we can use new classes and features. 
- If any not java8 compatible feature was used in the main code, compilation will fail
- artipie-main, hex and maven adapters can use java 21 in the main code (these modules are not used by Chinees team)

### DIFF
--- a/.github/workflows/maven-adapter-release.yml
+++ b/.github/workflows/maven-adapter-release.yml
@@ -2,7 +2,7 @@ name: Create Maven adapter release
 on:
   push:
     tags:
-      - '*_*_*'
+      - '*_*'
 jobs:
   build:
     name: Build release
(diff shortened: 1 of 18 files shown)


olenagerasimova:
@dgarus please, take a look

dgarus (approved):
Thanks!

PR #1351, opened 2023-12-04
Trying to hot-fix s3 bug. Hangs in reads from asto

ChGen:
This is hotfix for conda+s3 bug.  There are two interrelated issues (parts) in this bug:
1. Hangs in reads from asto ContentAsInputStream.inputStream().
Instances of PublisherAs(content) can read data from content correctly, so modified implementation is introduced, which uses `byte[]` instead of `InputStream`. Also, anaconda may hang up after some time after hang, but may not print any error messages. This is hot-fix, so deeper bug investigation will be performed after it. 
2. In case of the file stuck in `.uploads/` directory (broken connection, artipie freeze, reboos, client quits, etc.), adapter just returns BAD_REQUEST without any extra log messages or `.uploads/` cleanup. Moreover, anaconda doesn't print any error messages. As for now, I use `Storage.exclusively()` for synchronizing uploads storage access. It provides explicit synchronization and sometimes better error messages in artipie logs or anaconda output. More discussion will be required for possibly better fix.

### DIFF
--- a/conda-adapter/src/main/java/com/artipie/conda/asto/AstoMergedJson.java
+++ b/conda-adapter/src/main/java/com/artipie/conda/asto/AstoMergedJson.java
@@ -5,15 +5,24 @@
 package com.artipie.conda.asto;
 
 import com.artipie.asto.ArtipieIOException;
+import com.artipie.asto.Content;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
+import com.artipie.asto.ext.PublisherAs;
 import com.artipie.asto.misc.UncheckedIOFunc;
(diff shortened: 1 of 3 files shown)


ChGen:
@olenagerasimova , please review, I moved bug hotfix to `conda-adapter`. I've tried also calling explicitly `close()` on parser, but it doesn't help.

dgarus:
@ChGen 
Why do you think these changes are fixed the bug?

ChGen:
@dgarus , I can reproduce both parts of the bug locally and test that hotfix works. As a next step, I'll provide extra IT tests for asto+s3 too.

dgarus (approved):

PR #1356, opened 2023-12-15
feat: removed ppom usage

olenagerasimova:
Removed ppom, moved profiles and plugins into common pom


olenagerasimova:
@dgarus please, check

dgarus (approved):
Thanks!

PR #1370, opened 2024-01-10
fix: correct release scripts and description

### LINKED ISSUE #1364: Release java version for adapters should be 17 now
Let's update java release version to 17 as our colleagues requested. 
Also, maven release should release all modules (except from artipie-main) at the same time.

olenagerasimova:
closes #1364


dgarus (approved):
thanks

PR #1372, opened 2024-01-11
deps: update vulnerable guava

olenagerasimova:
Updated [vulnerable guava](https://mvnrepository.com/artifact/com.google.guava/guava/32.0.0-jre) version


dgarus:
Can we define a property with guava version in the root pom file and use it in submodules?

olenagerasimova:
@dgarus sure, thanks

dgarus (approved):
thanks

Issue #1343, opened 2023-11-29
Broken uploads on Conda with S3 storage


ChGen:
Broken upload operations for conda repositories with S3 storage. Looks like filesystem storage isn't affected. Could be reproduced with local instance of `artipie` with local `minIO` S3 server.  It works mostly for the first package upload, subsequent uploads are getting struck in `.upload` folder.

Also:
* amazon s3 client should be updated to the new version. 
* asto-core should be buildable & testable with jdk21
* more integration/smoke tests with s3 storage servers required

See also logs below:

[Full logs.zip](https://github.com/artipie/artipie/files/13501499/logs.zip)

```
Uploading file "aiteam/scipy/1.9.3/linux-64/scipy-1.9.3-py310hdfbd76f_2.tar.bz2"
26.2MB [00:04, 6.17MB/s]
java.util.concurrent.CompletionException: software.amazon.awssdk.services.s3.model.S3Exception: You must provide the Content-Length HTTP header. (Service: S3, Status Code: 411, Request ID: C4PMMMQ7Y57S90N7, Extended Request ID: 2Hkds1PSqtfYKGb/PguPDXoKAbCBFJu/kN5K0/VuPyrCp7MH7Ezvhe86ZWQQpwh/sbh1U1hal44=)
```

```
.....
Ram, [24.11.2023 17:33]
INFO: Exists 'conda/linux-64/repodata.json': false
Nov 24, 2023 2:26:07 PM com.artipie.asto.LoggingStorage log
INFO: Save 'conda/linux-64/repodata.json': Optional.empty
Nov 24, 2023 2:26:08 PM com.artipie.asto.LoggingStorage log
INFO: Move 'conda/.upload/linux-64/ai-utils-0.2.0-py310_0.tar.bz2' 'conda/linux-64/ai-utils-0.2.0-py310_0.tar.bz2'

Ram, [24.11.2023 17:34]
here when Exists 'conda/linux-64/repodata.json': false
it worked

Ram, [24.11.2023 17:35]
there were no errors

Ram, [24.11.2023 17:35]
It didn't complain about anything in case of success scenario

Ram, [24.11.2023 17:40]
If I tried to restart and re-login in to the artipie server between the uploads, it is not hanging, but it is throwing like this

Ram, [24.11.2023 17:40]
bade02r@MAC-Y42VW654JQ Downloads % anaconda upload ai-utils-0.2.0-py310_0.tar.bz2
Using Anaconda API: http://localhost:8080/conda
[...]

PR #1373, opened 2024-01-12
Conda + s3 bug fix

ChGen:
Fixing deadlock in `StorageValuePipeline.java` + test in asto-s3. Hotfix is removed.
Deadlock happens when Publisher for the `InputStream` produces data in the same thread as `action` callback by several chunks, and `action` callback actually does blocking `read()` on provided `InputStream`. Since plain `read()` is blocking, event loop doesn't get called, Publisher doesn't provide next part of the data and `read()` gets stuck. This situation always happens for S3Storage, since AWS S3 SDK invokes callback in one of its AWS threads and provides Publisher from that same thread too. The solution was to call `action` with the blocking `read()` in async block, from different thread.
Also, I had problems with `PublishingOutputStream` being stuck or incorretly callbacked, if `onComplete()` wasn't called in time.

Part of https://github.com/artipie/artipie/issues/1343

### DIFF
--- a/artipie-main/examples/run.sh
+++ b/artipie-main/examples/run.sh
@@ -182,6 +182,11 @@ r=0
 grep "FAILED" "$workdir/results.txt" > /dev/null || r="$?"
 if [ "$r" -eq 0 ] ; then
   rm -fv "$pidfile"
+  echo "Artipie container logs:"
+  container=$(docker ps --filter name=artipie -q 2> /dev/null)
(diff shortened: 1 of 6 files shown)


ChGen:
@olenagerasimova , @dgarus , please review.

ChGen:
@dgarus , please review, I added test with mixed read/write operations and ReadWithDelaysStorage on top of the S3Storage.

dgarus (approved):

PR #1374, opened 2024-01-12
fix: removed unused servlet slice

olenagerasimova:
Removed unused servlet slice and cleaned up artipie-core pom.


dgarus (approved):
Thanks!

PR #1377, opened 2024-01-16
Helm tests fix

ChGen:
While working on https://github.com/artipie/artipie/issues/1343 I had intermittent helm adapter tests fails. So I investigated it and prepared helm adapter tests fix. The problem was that java `java.io.BufferedWriter` doesn't propagate `close()` call downstream, it just flushes its own data and resets its reference. So, closing streams properly allowed also to enable few other disabled helm tests without intermittent test fails, as yet.
Interestingly, in [JDK7](https://github.com/openjdk/jdk/blob/jdk7-b147/jdk/src/share/classes/java/io/BufferedWriter.java#L258) it would close its stream, but in [JDK8](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/io/BufferedWriter.java#L259) it didn't. [JDK21](https://github.com/openjdk/jdk/blob/jdk-21%2B35/src/java.base/share/classes/java/io/BufferedWriter.java#L375) (last LTS) doesn't close its stream too.

### DIFF
--- a/helm-adapter/src/main/java/com/artipie/helm/AddWriter.java
+++ b/helm-adapter/src/main/java/com/artipie/helm/AddWriter.java
@@ -122,9 +122,8 @@ interface AddWriter {
             return CompletableFuture.allOf().thenCompose(
                 none -> {
                     try {
-                        final BufferedWriter bufw = new BufferedWriter(
-                            new OutputStreamWriter(Files.newOutputStream(out))
-                        );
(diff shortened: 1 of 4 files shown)


ChGen:
@olenagerasimova , @dgarus , please review.

olenagerasimova (approved):
Thanks!

Issue #1375, opened 2024-01-12
Artipie behind a proxy


vyakovlev-hw:
Hello All,

My `artipie` installation is behind a corporate proxy, so `artipie` can NOT connect to any site, e.g. `registry-1.docker.io` without the proxy.

Consider this diagram

![test drawio](https://github.com/artipie/artipie/assets/62640940/a0a49268-263c-44e1-bf50-ac6b16278536)

A user wants to get an `nginx` image from `artipie`, that is set up to forward requests to `registry-1.docker.io` (described in [this](https://github.com/artipie/artipie/blob/master/.wiki/repositories/docker-proxy.md) article). The server tries to fetch the image from `registry-1.docker.io` and fails with:

<details>
<summary><b>errors</b></summary>
[vert.x-eventloop-thread-0] INFO com.artipie.http.MainSlice - >> HEAD /v2/dockerhub/nginx/manifests/latest HTTP_1_1

Host: host:8080
User-Agent: docker/20.10.23 go/go1.18.10 git-commit/6051f14 kernel/5.4.0-128-generic os/linux arch/amd64 UpstreamClient(Docker-Client/20.10.23 \(linux\))
Accept: application/vnd.docker.distribution.manifest.v2+json
Accept: application/vnd.docker.distribution.manifest.list.v2+json
Accept: application/vnd.oci.image.index.v1+json
Accept: application/vnd.docker.distribution.manifest.v1+prettyjws
Accept: application/json
Accept: application/vnd.oci.image.manifest.v1+json
Authorization: Basic xxx==
Connection: close
[vert.x-eventloop-thread-0] INFO com.artipie.settings.cache.CachedUsers - Successfully authenticated 'docker' user via CachedUsers(size=1),origin=Joined([Joined([AuthFromEnv(),GithubAuth()]),AuthFromStorage])
Jan 12, 2024 2:40:51 PM com.artipie.asto.LoggingStorage log
INFO: Exists 'dockerhub/docker/registry/v2/repositories/nginx/_manifests/tags/latest/current/link': false
[HttpClient@604ed9f0-scheduler-1] ERROR com.artipie.docker.composite.MultiReadManifests - Failed to read manifest latest: java.util.concurrent.CompletionException: java.net.SocketTimeoutException: Connect Timeout
        at java.base/java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:368)
[...]

PR #1385, opened 2024-01-25
feat: Http client settings for xxx-proxy adapters

dgarus:
Close: #1375 

- ability to define `http` client setting in `meta` and `repo` configs
- ability to define several proxy servers 
- using of `https.proxyHost` property


artipie-main/src/main/java/com/artipie/SliceFromConfig.java
```diff
@@ -307,4 +310,10 @@ private static Slice build(
             )
             .get();
     }
+
+    private static ClientSlices clientSlices(final RepoConfig cfg, final Settings settings) {
+        return new JettyClientSlices(
```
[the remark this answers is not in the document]
olenagerasimova: Also, should not `JettyClientSlices` it be started before usage?

olenagerasimova (commented):
@dgarus please, take a look: some typos and important question

artipie-main/src/main/java/com/artipie/SliceFromConfig.java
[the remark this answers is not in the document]
dgarus: @olenagerasimova thanks, a good point.

dgarus:
@olenagerasimova 
I added `SlicesCache` that manages `JettyClientSlices`.
Please take a look.

artipie-main/src/main/java/com/artipie/RepositorySlices.java
```diff
@@ -0,0 +1,131 @@
...
+     * @return Slice for repo
+     */
+    private SliceValue resolve(final Key name, final int port) {
+        final RepoConfig cfg = new RepositoriesFromStorage(this.settings)
+            .config(name.string()).toCompletableFuture().join();
+        if (cfg.port().isEmpty() || cfg.port().getAsInt() == port) {
+            final JettyClientSlices client = new JettyClientSlices(
+                cfg.httpClientSettings().orElseGet(settings::httpClientSettings)
+            );
+            client.start();
```
olenagerasimova, line 97: hm, http client is required for proxy repository only, so why create it for each repository?

olenagerasimova (commented):
great, although I have one question

artipie-main/src/main/java/com/artipie/RepositorySlices.java
dgarus: A good point!
I'm going to implement a more suitable approach in #1386

olenagerasimova: for now, it's necessary to make sure all the integration tests in artipie pass... There are tests with proxies, so we'll know for sure if smth is broken

olenagerasimova (approved):

Issue #1383, opened 2024-01-24
Bug: PyPI repository upload to S3 storage is broken


EDF-MarcellO:
I'm running Artipie using AWS App Runner in a containerized deployment with S3 storage specified. When I want to upload PyPI packages to the PyPI repository hosted on Artipie, I am greeted with following error:

Upload command:
python3 -m twine upload --repository-url https://domain.region.awsapprunner.com/pypi -u username -p password --verbose numpy-1.26.3-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
(package available from: https://pypi.org/project/numpy/#files)

Response:
INFO Response from https://domain.region.awsapprunner.com/pypi:                   
         400 Bad Request                                                                     
INFO Bad request                                                                         
         software.amazon.awssdk.services.s3.model.S3Exception: You must provide the          
         Content-Length HTTP header. (Service: S3, Status Code: 411, Request ID:             
         SomeRandomID, Extended Request ID: AnotherRandomID)

What I would like to see:
Successful upload of wheel file to Artipie PyPI repository.

N.B.: Placing the wheel file into the dist folder and running an updated command (to upload all files in dist folder, dist/*) yields the same results.

Repository configuration yaml file:
repo:
  type: pypi
  storage:
    type: s3
    bucket: my-s3-bucket
    region: my-s3-region
    
(obviously, some domain names have been redacted for security purposes, happy to share details privately)

bade02r:
Hi @ChGen,  is this fix released with in the latest version? if not, when are you planning to release?

bade02r:
@ChGen , thanks for the fix, would you mind to release a minor version with the latest fixes?

ChGen:
@bade02r , published `artipie/artipie:v0.30.13`. Feedback is appreciated :)

ChGen:
Should be fixed in `artipie/artipie:v0.30.14`. Please, reopen with fresh logs in case of errors.

Issue #1368, opened 2024-01-09
rpm repository - error uploading my rpm artifact to S3


Tzahif:
Hi,
I am trying to upload my rpm artifact to  S3 and I am getting an error.
Uploading the same artifacts to FS is working

Repo config:
my-rpm-s3.yaml
````
repo:
  type: rpm
  storage:
    type: s3
    bucket: repo-test
    region: us-east-1
  layout: flat
  settings:
    digest: sha256 # packages digest algorithm
    naming-policy: sha1 # naming policy for metadata files
    filelists: true # is filelist metadata file required
    # repository update mode:
    update:
      # update metadata on package upload
      on: upload
      # or schedule the update
#      on:
#        cron: 0 2 * * *
````

Upload to S3 command and response:
````
➜  rpm curl -i -X PUT --data-binary "@my-package.rpm" http://0.0.0.0:8888/my-rpm-s3/my-package.rpm
HTTP/1.1 100 Continue

HTTP/1.1 500 Internal Server Error
content-length: 11535

java.util.concurrent.CompletionException: software.amazon.awssdk.services.s3.model.S3Exception: You must provide the Content-Length HTTP header. (Service: S3, Status Code: 411, Request ID: 29EGX35GYMM39Q1R, Extended Request ID: EkC/y3ABxoVT/R+EDM8BZr9vli2RlaJhygCDN0O6WhN6ENpBvkbxKpSITCwB0gKgG+FSb8+BW14=)
java.util.concurrent.CompletionException: software.amazon.awssdk.services.s3.model.S3Exception: You must provide the Content-Length HTTP header. (Service: S3, Status Code: 411, Request ID: 29EGX35GYMM39Q1R, Extended Request ID: EkC/y3ABxoVT/R+EDM8BZr9vli2RlaJhygCDN0O6WhN6ENpBvkbxKpSITCwB0gKgG+FSb8+BW14=)
        at software.amazon.awssdk.utils.CompletableFutureUtils.errorAsCompletionException(CompletableFutureUtils.java:62)
        at software.amazon.awssdk.core.internal.http.pipeline.stages.AsyncExecutionFailureExceptionReportingStage.lambda$execute$0(AsyncExecutionFailureExceptionReportingStage.java:51)
        at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:934)
        at java.base/java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:911)
[...]

Issue #1367, opened 2024-01-09
debian repository - error uploading my deb artifact to S3


Tzahif:
Hi,
I am trying to upload my deb artifact to S3 and I am getting errors.
Repos configs:

my-deb-s3.yaml
````
repo:
  type: deb
  storage:
    type: s3
    bucket: repo-test
    region: us-east-1
  settings:
    Components: main
    Architectures: amd64 arm64
````

Upload to S3 command and response:
````
➜  debian curl -i  -X PUT  --data-binary "@my-package.deb" http://0.0.0.0:8888/my-deb-s3/main/my-package.deb

HTTP/1.1 100 Continue

HTTP/1.1 500 Internal Server Error
content-length: 5935

java.util.concurrent.CompletionException: com.artipie.asto.ArtipieIOException: java.io.IOException: Key does not exist: my-deb-s3/main/my-package.deb
java.util.concurrent.CompletionException: com.artipie.asto.ArtipieIOException: java.io.IOException: Key does not exist: my-deb-s3/main/my-package.deb
        at java.base/java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:368)
        at java.base/java.util.concurrent.CompletableFuture.completeRelay(CompletableFuture.java:377)
        at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1152)
        at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
        at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2179)
        at com.artipie.asto.s3.S3Storage.lambda$exists$0(S3Storage.java:122)
        at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:934)
        at java.base/java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:911)
        at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
        at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2194)
        at software.amazon.awssdk.utils.CompletableFutureUtils.lambda$forwardExceptionTo$0(CompletableFutureUtils.java:76)
        at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:863)
[...]

PR #1400, opened 2024-02-13
Updating s3 and jetty deps

ChGen:
Updating s3 and jetty deps. Reduce potential conflicts over jetty, vulnerabilities and AWS compatibility issues.

### DIFF
--- a/debian-adapter/pom.xml
+++ b/debian-adapter/pom.xml
@@ -104,13 +104,13 @@ SOFTWARE.
     <dependency>
       <groupId>com.adobe.testing</groupId>
       <artifactId>s3mock</artifactId>
-      <version>3.4.0</version>
+      <version>${s3mock.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.adobe.testing</groupId>
       <artifactId>s3mock-junit5</artifactId>
-      <version>3.4.0</version>
+      <version>${s3mock.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/debian-adapter/pom.xml
+++ b/debian-adapter/pom.xml
@@ -104,25 +104,13 @@ SOFTWARE.
     <dependency>
       <groupId>com.adobe.testing</groupId>
       <artifactId>s3mock</artifactId>
-      <version>2.4.9</version>
+      <version>3.4.0</version>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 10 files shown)


ChGen:
@dgarus , please review.

debian-adapter/pom.xml
```diff
@@ -104,25 +104,13 @@ SOFTWARE.
     <dependency>
       <groupId>com.adobe.testing</groupId>
       <artifactId>s3mock</artifactId>
-      <version>2.4.9</version>
+      <version>3.4.0</version>
```
dgarus, line 107: We should create a property for `s3mock`'s version in the parent pom

dgarus (changes requested):
please check my comment

debian-adapter/pom.xml
ChGen: @dgarus , I can make local variable, like for `jettyVersion`. For some reason `jettyVersion` is not inherited from parent pom - without local declaration maven does not proceed. Also IDE Idea 2022 & 2023 with WSL2 makes projects with submodules hardly usable due to several bugs, so I open every module as separate project and need local declaration.

dgarus: I don't like an approach when you have to change several places to update one artifact version.
Look at the `guava.version` property as an aexample.

ChGen: @dgarus , ok, I'll check my environment, maybe also something broken in it.

Tzahif:
@ChGen, because i raised the issues with S3, [debian](https://github.com/artipie/artipie/issues/1367), [rpm](https://github.com/artipie/artipie/issues/1368) and [docker](https://github.com/artipie/artipie/issues/1369), i can help with testing in my envs (minio & real S3 bucket).
i built this fix branch from source and i still get the same errors.
the only thing that works is rpm over minio (not real s3).
debian and docker still doesn't work with this fix, both minio and real S3 bucket.

ChGen:
@Tzahif , thanks for reporting, I'll continue with https://github.com/artipie/artipie/issues/1367 using minIO configuration you described. And with this changes too.

ChGen:
@dgarus , please review.

dgarus (approved):

ChGen:
Affected issues:
https://github.com/artipie/artipie/issues/1383
https://github.com/artipie/artipie/issues/1368
https://github.com/artipie/artipie/issues/1367

PR #1416, opened 2024-03-04
feat: extended content interface

### LINKED ISSUE #1415: Add to the Content interface a set of getting data methods in different formats
Currently, we have plenty of boilerplate code to get the `Content` data as a `String` or as a `byte[]`. 
It will be very convenient to have according methods which return the `Content` data directly from the instance of this interface.

So we can add the following methods:
```CompletableFuture<String> asStringFuture()```
```String asString()```
```CompletableFuture<byte[]> asBytesFuture()```
```byte[] asBytes()```

dgarus:
Closes #1415 
- added a set of getting data methods in different formats
- removed `PublisherAs` classes


dgarus:
@ChGen hi! please review

ChGen (approved):
@dgarus , thanks!

Issue #1409, opened 2024-02-23
The test RpmSliceS3ITCase is broken


dgarus:
The test failed with the following error

```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'fileStoreController' defined in class path resource [com/adobe/testing/s3mock/S3MockConfiguration.class]: Unsatisfied dependency expressed through method 'fileStoreController' parameter 0: Error creating bean with name 'objectService' defined in class path resource [com/adobe/testing/s3mock/service/ServiceConfiguration.class]: Unsatisfied dependency expressed through method 'objectService' parameter 1: Error creating bean with name 'objectStore' defined in class path resource [com/adobe/testing/s3mock/store/StoreConfiguration.class]: Failed to instantiate [com.adobe.testing.s3mock.store.ObjectStore]: Factory method 'objectStore' threw exception with message: jakarta/xml/bind/JAXBException

	at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:798)
	at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:542)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1334)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1164)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:561)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:521)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:325)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:323)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:199)
[...]

PR #1418, opened 2024-03-06
Test was failing due to dependency issues

ChGen:
Test was failing due to dependency issues: https://github.com/artipie/artipie/issues/1409


ChGen:
@dgarus , please review.

PR #1434, opened 2024-03-26
Developer information must be in root pom.xml for publishing artifacts.

ChGen:
Developer information must be in root pom.xml for publishing artifacts.


ChGen:
@dgarus , please review.

dgarus (approved):

PR #1440, opened 2024-04-05
Fix possible crash

ChGen:
Fix possible crash. Backported fix from `origin/jdk17-builds`


ChGen:
@dgarus , please review.

dgarus (approved):

PR #1455, opened 2024-05-08
Testing parallel IT tests runs

ChGen:


ChGen:
Parallel integration tests may provide extra 1.5x speed-up, but requires a lot of refactoring to be stable, so it’s not viable. Previous docker caching already provided 2x speed-up (60min to 30min) for integration tests.

PR #1478, opened 2025-02-24
ASTO S3 - added aws sts dependency

Tzahifer:
ASTO S3 - added aws sts dependency


Tzahifer:
@ChGen i would appreciate your review here

ChGen (approved):

ChGen:
Thanks!

